PR #3, opened 2019-09-04
Build-related update

mhanus:
See the commit messages for more details

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -13,7 +13,7 @@ else()
     add_definitions(-DWINDOWS_ENV)
 endif()
 
-#------------------------------------------------ DEPENDECIES
+#------------------------------------------------ DEPENDENCIES
 if (NOT (DEFINED ENV{BOOST_ROOT}))
     message(FATAL_ERROR "***** BOOST_ROOT is not set *****")
 else()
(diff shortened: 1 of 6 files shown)


CMakeLists.txt
```diff
@@ -100,26 +80,43 @@ include_directories("${PROJECT_SOURCE_DIR}/CHI_TECH")
...
-
 set(TRIANGLE "${TRIANGLE_ROOT}/triangle.o")
 
-set(LIBS lua m dl ${MPI_CXX_LIBRARIES} petsc)
-set(LIBS ${LIBS} vtkCommonCore${VTK_VERSION} vtkCommonDataModel${VTK_VERSION})
-set(LIBS ${LIBS} vtkIOLegacy${VTK_VERSION} vtkIOCore${VTK_VERSION})
-set(LIBS ${LIBS} vtkIOXML${VTK_VERSION} vtkParallelCore${VTK_VERSION} )
-set(LIBS ${LIBS} vtkIOParallelXML${VTK_VERSION})
+# --------------------------- VTK
+find_package(VTK COMPONENTS
```
Naktakala: This seems to work fine on local machines with VTK installed. Just for #shits&giggles I tested your fork on quartz and then it fails without really any message on how to fix it.

If we add "REQUIRED": `find_package(VTK REQUIRED COMPONENTS` then add least cmake fails and tells the user to set `VTK_DIR` after which it works.

mhanus: Good point - I've pushed the update.

PR #9, opened 2019-09-10
Cleanup of project directory

Naktakala:
This is a small cleanup. I've moved the cross-sections to `CHI_RESOURCES/CrossSections` and updated the regression tests.


Naktakala:
Yeah I'd be fine with that. There is really only 2 or 3 cross-sections used for the tests. Eventually I do want to supply a data-library which will to `CrossSections`.

Naktakala:
Done. The core test cross-sections have been added to the test directory and the test modified accordingly.

mhanus:
Looks good to me now. 

One thing to consider as a possible feature improvement: `chiPhysicsMaterialSetProperty` currently assumes the data path to be relative to the cwd (in the case of the tests, the root chi-tech directory). If I executed ChiTech from a different folder, I'd have to update all `chiPhysicsMaterialSetProperty` calls. Wouldn't it be better to first search in the folder of the input .lua file itself, and if not found, try cwd?

PR #20, opened 2019-10-28
Build system script

Naktakala:
This is the first rendition of a framework builder which installs all the dependencies for ChiTech. It's been tested on WSL running ubuntu and general ubuntu systems and should alleviate most of troubles people expressed in complaints.

Tested on:
- WSL running ubuntu
- ubuntu 18.04 and 19.04
- CentOS 7
- MacOS Mojave (using gcc 9.2)


CHI_DOC/Install_ubuntu_easy.md
```diff
@@ -0,0 +1,157 @@
...
+
+If this is not displayed then install gcc using the following command:
+
+```bash
+sudo apt-get install build-essential gfortran
+```
+
+If you still do not get the appropriate version when running either ``gcc --version``
+or ``gfortran --version`` then there are many online resources which may be able to
+assist you. One issue commonly seen is that the repositories on your system is not
```
mhanus: repositories on your system **are** not updated

CHI_DOC/Install_ubuntu_easy.md
```diff
@@ -0,0 +1,157 @@
+# Easy install on Ubuntu Machines
+
+The following instructions were tested on Ubuntu 18.04 LTS; other Linux
+distributions might require some minor tweaking.
+
+### Step 1 - Installing GCC, GFortran and the basic environment
+
+GCC is used to build and install ChiTech.
+GFortran and Python is used during the installation of PETSc
```
mhanus: GFortran and Python **are** used

CHI_RESOURCES/configure_dependencies.py
```diff
@@ -0,0 +1,506 @@
...
+    os.chdir("readline-8.0")
+    env_vars=os.environ.copy()
+    command = "./configure --prefix=" + \
+              install_dir + \
+              "/READLINE/readline-8.0/build"
+    success,err = ExecSub(command,log_file,env_vars)
+    if (not success):
+      print(command,err)
+
+    command = "make -j8"
```
mhanus, line 226: I think it might be useful (here and at other instances of `make -j8`) to allow an additional optional argument to specify the number of parallel makes (you can keep the default 8 if not specified). For people with older machines (like me :-)), overloading the machine might really slow down the whole build process of the larger dependencies (e.g. VTK), and on the other hand, I can imagine people wanting to use more

mhanus (approved):

CHI_RESOURCES/configure_dependencies.py
Naktakala: I have another update planned and it is something I also identified as a feature. Will include it there

README.md
```diff
@@ -5,6 +5,14 @@ as part of a research study.
 
 ## Installation instructions
 
+We have a set of easy instructions for users running Ubuntu 18.04+ (including WSL
+with Ubuntu 18.04)
+
+Easy scripts - [Easy Linux instructions](CHI_DOC/Install_ubuntu_easy.md)
+
+If the automated installation of dependencies fail, or if you don't have a system
```
ragusa: fails

README.md
```diff
@@ -5,6 +5,14 @@ as part of a research study.
...
+supporting it, then rather follow any of the below instructions:
```
ragusa: --> then follow the instructions below:

CHI_DOC/Install_ubuntu_easy.md
```diff
@@ -0,0 +1,157 @@
...
+
+In the main directory (i.e. *chi-tech/*), execute:
+```bash
+    $ make -j4
+```
+
+You can also use -j8 even if you don't have 8 processors, the make command
+will use threading where possible.
+
+Whenever you need to recompile and the configuration might have changed
```
ragusa: Whenever recompilation is needed or configuration has changed, ....

CHI_DOC/Install_ubuntu_easy.md
```diff
@@ -0,0 +1,157 @@
...
+```
+
+To check if this is working properly, simply do the mpi version of checking for
+a C++ compiler:
+
+```bash
+mpicc --version
+```
+
+Which should display the same message the gcc call did, i.e.
```
ragusa: i.e.,

CHI_DOC/Install_linux.md
```diff
@@ -298,7 +298,16 @@ In the main directory (i.e. *chi-tech/*), execute:
 You can also use -j8 even if you don't have 8 processors, the make command
 will use threading where possible.
 
-### Step 11 - ChiTech documentation
+### Step 11 - Run regression tests
+
+To check if the code compiled correctly execute the test scripts:
```
ragusa: add comma before execute

ragusa (changes requested):

README.md
Naktakala: Done

README.md
Naktakala: Done

CHI_DOC/Install_linux.md
Naktakala: Done

CHI_DOC/Install_ubuntu_easy.md
Naktakala: Done

README.md
```diff
@@ -5,6 +5,14 @@ as part of a research study.
 
 ## Installation instructions
 
+We have a set of easy instructions for users running Ubuntu 18.04+ (including WSL
```
wdhawkins, line 8: It might be a good idea at this point to have the user check to make sure sudo will work.  Depending on their luck, we can point them in the right direction.

Naktakala: Added statements to install instructions.

wdhawkins (approved):

PR #21, opened 2019-10-29
Gauss leg rootfind pr

Naktakala:
For some reason this escaped me during a port-over but here it is fixed now. The root finding algorithm for finding the roots of the Legendre polynomials is now up to par with the algorithm used in Golub-Fischer.

This algorithm allows us to find the abscissae for a quadrature in excess of 2000 angles.


CHI_TECH/ChiMath/Quadratures/LegendrePoly/legendrepoly.cc
```diff
@@ -38,8 +40,30 @@ double chi_math::dLegendredx(int N, double x)
...
 
   return retval;
-}
\ No newline at end of file
+}
+
+//###################################################################
+/**Provides the function evaluation of the second derivative of Pn at value x
+
+ \param N int The Legendre polynomial.
```
mhanus: This is misleading - "Order of the Legendre polynomial" would read better

CHI_TECH/ChiMath/Quadratures/quadrature_gausslegendre.cc
```diff
@@ -12,64 +20,125 @@ void chi_math::QuadratureGaussLegendre::
...
+      << "roots that it might just be prudent to not use a quadrature.";
   }
-  //printf("Number of possible roots=%d\n",abscissae.size());
 
-  //============================================= Newton iteration to find root
-  for (unsigned k=0; k<abscissae.size(); k++)
+  // For this code we simply check to see where the
+  // polynomial changes sign.
+  double delta = 2.0/num_search_intvls;
+  std::vector<double> xn(N, 0.0);
```
mhanus: Is the reason why you do everything on `xn` and at the end move to `roots` that you wanted to make the code more readable (more similar to the math text)? If so, a better and more efficient way would be just to define an alias for `roots`:
```c++
std::vector<double>& xn = roots;
xn.resize(N, 0.0);
...
```

CHI_TECH/ChiMath/Quadratures/quadrature_gausslegendre.cc
```diff
@@ -12,64 +20,125 @@ void chi_math::QuadratureGaussLegendre::
...
+  if (N>256)
+    num_search_intvls *= 10;
+  if (N>768)
+    num_search_intvls *= 10;
+
+  if (N>2056)
+  {
+    num_search_intvls *= 10;
+    chi_log.Log(LOG_0WARNING)
+      << "chi_math::QuadratureGaussLegendre::FindRoots: "
```
mhanus, line 79: This reads weird... 
"Accuracy of the root finder will be diminished along with a reduction in accuracy"  – what is the second *accuracy*?
"At this point there are so many roots that it might just be prudent to not use a quadrature" – what else to use then? Monte Carlo integration?

CHI_TECH/ChiMath/Quadratures/quadrature_gausslegendre.cc
```diff
@@ -12,64 +20,125 @@ void chi_math::QuadratureGaussLegendre::
...
+      //===================== Compute final formula
+      double a    = fpp + f*S2;
+      double b    = B_xk*B_xk + fp*fp - f*a;
+      double c    = 2.0*f*B_xk/b;
+
+      xn[k] = xold - c;
+
+      if (std::fabs(xn[k]-xold) < tol)
+        break;
     }//while
```
mhanus: there is no *while* – either remove the comment, or say something like `// Newton's iteration loop`

CHI_TECH/ChiMath/Quadratures/quadrature_gausslegendre.cc
```diff
@@ -12,64 +20,125 @@ void chi_math::QuadratureGaussLegendre::
...
+      Legendre(N+1, abscissae[k]) /
+      Legendre(N+1, abscissae[k]);
 
-    xgold = xgnew;
-    fold = fnew;
+    if (verbose)
+      chi_log.Log(LOG_0)
+        << "root[" << k << "]=" << abscissae[k]
+        << ", weight=" << weights[k];
+  }//for root
```
mhanus: The comment ending the loop doesn't match the loop definition

CHI_TECH/ChiMath/Quadratures/quadrature_gausslegendre.cc
```diff
@@ -12,64 +20,125 @@ void chi_math::QuadratureGaussLegendre::
...
   {
-    xgnew = xgold + dx;
-    fnew = Legendre(N,xgnew);
-    //printf("xgnew=%f Fnew=%f  Fold=%f\n",xgnew,fnew,fold);
-    if ((fnew*fold)<0.0)
-    {
-      abscissae.push_back(xgnew);
-      weights.push_back(1.0);
-    }
+    weights[k] =
```
mhanus, line 31: IMO, `f = a/(b*c*d*e)` is more understandable than `f = a/b/c/d/e` and it could also be a tiny bit faster (multiplication should be generally faster than division, see e.g. [here](https://stackoverflow.com/questions/4125033/floating-point-division-vs-floating-point-multiplication/5322101)). I suggest rewriting using multiplication, but if other reviewers are fine with that and you want to keep the divisions, I won't fight for it :-) 

As a side note - I became curious, and since I was reading this right after today's 689 class where I was doing some benchmarking, I wrote a small quick [test](http://quick-bench.com/ZvhKG5wQsG34fKt1V0T-jh7gzGM)... you'll see that the performance is probably not what we should be concerned about here, but for me it is really the readability...

mhanus (changes requested):

CHI_TECH/ChiMath/Quadratures/quadrature_gausslegendre.cc
Naktakala: Second "accuracy" is supposed to be "stability".

Naktakala: Also, removed the last sentence.

CHI_TECH/ChiMath/Quadratures/quadrature_gausslegendre.cc
Naktakala: Yes, good point. The `xn` is also now changed to `xk` which aligns more with the paper.

CHI_TECH/ChiMath/Quadratures/quadrature_gausslegendre.cc
Naktakala: Good catch!

mhanus (approved):

CHI_TECH/ChiMath/Quadratures/LegendrePoly/legendrepoly.cc
```diff
@@ -1,10 +1,12 @@
 #include "legendrepoly.h"
 #include <math.h>
+
```
wdhawkins, line 3: Even though it's not really part of this diff, please change
`#include<math.h>`
to
`#include<cmath>`

Naktakala: Done. Thanks! It still exists some places and I will eliminate it where I find it.

CHI_TECH/ChiMath/Quadratures/quadrature_gausslegendre.cc
```diff
@@ -12,64 +20,123 @@ void chi_math::QuadratureGaussLegendre::
...
+ * \param N Is the order of the polynomial.
+ * \param roots Is a reference to the roots.
+ * \param max_iters Maximum newton iterations to perform for each root.
+ *        Default: 1000.
+ * \param tol Tolerance at which the newton iteration will be terminated.
+ *        Default: 1.0e-12.
+ *
+ * \author Jan*/
+void chi_math::QuadratureGaussLegendre::FindRoots(
+  int N, std::vector<double> &roots, size_t max_iters, double tol)
```
wdhawkins: Why is max_iters of type size_t instead of int?

CHI_TECH/ChiMath/Quadratures/quadrature_gausslegendre.cc
```diff
@@ -12,64 +20,123 @@ void chi_math::QuadratureGaussLegendre::
...
+{
+  //======================================== Populate init guess
+  //This initial guess proved to be quite important
+  //at higher N since the roots start to get
+  //squeezed to -1 and 1.
+  size_t num_search_intvls = 1000;
```
wdhawkins: In general, I would suggest we avoid size_t unless we know we need it.

CHI_TECH/ChiMath/Quadratures/quadrature_gausslegendre.cc
Naktakala: IDK. Can't remember? Why not? Maybe cause it should be >=0. What would you like it to be? int ..? done

wdhawkins: Also, I think it's preferable to have:

`std::vector<double> chi_math::QuadratureGaussLegendre::FindRoots(
    int N, size_t max_iters, double tol)`

instead of passing in roots as a parameter.  It's cleaner IMHO and RVO should make it just as speedy.


`

Naktakala: I like that more too. Let's do that.

wdhawkins: > IDK. Can't remember? Why not? Maybe cause it should be >=0. What would you like it to be? int ..? done

This is something we should probably discuss and include in our style guide.  I tend to try to avoid unsigned integers if possible (unless we're using them in the context of addresses or sizes). I think it creates more problems than it solves.  This sort of sums it up for me: https://wesmckinney.com/blog/avoid-unsigned-integers/.

wdhawkins (approved):

PR #23, opened 2019-11-03
Removed the Eigen library as a dependency

Naktakala:
The removal of Eigen as a dependency was straight forward. I also took the chance to port over some math utils from PDT. Functionally nothing really changed except for energy collapsing for TGDSA


CHI_TECH/ChiMath/matrix_operations.cc
```diff
@@ -0,0 +1,512 @@
...
+  assert(r1 >= 0 && r1 < AR && r2 >= 0 && r2 < AR);
+
+  auto SwapVal = [](double& a, double& b)
+  {
+    double temp = a;
+    a = b;
+    b = temp;
+  };
+
+  for (unsigned int j = 0; j < AC; j++)
```
mhanus: No need for the lambda - just use 
```c++
std::swap(A[r1][j], A[r2][j]);
```

CHI_TECH/ChiMath/matrix_operations.cc
```diff
@@ -0,0 +1,512 @@
...
+    assert(c1 >= 0 && c1 < A[0].size() && c2 >= 0 && c2 < A[0].size());
+
+  auto SwapVal = [](double& a, double& b)
+  {
+    double temp = a;
+    a = b;
+    b = temp;
+  };
+
+  for (unsigned int i = 0; i < AR; i++)
```
mhanus: As above - `std::swap`

CHI_TECH/ChiMath/matrix_operations.cc
```diff
@@ -0,0 +1,512 @@
...
+  bool converged = false;
+  while(!converged && it_counter < max_it)
+  {
+    // Update old eigenvalue
+    lambda0 = lambda;
+    // Calculate new eigenvalue/eigenvector
+    Ay = MatMul(A, y);
+    lambda = Dot( y, Ay );
+    y = VecMul( Ay, 1./Vec2Norm(Ay) );
+    if (lambda < 0.)
```
mhanus: The sign flipping could be done after the convergence has been reached, just before returning -- although this is more a simplification of the algorithm rather than a significant performance boost.

CHI_TECH/ChiMesh/SurfaceMesher/Delaunay/delaunay_patch_00_utils.cc
```diff
@@ -1,5 +1,7 @@
 #include "delaunay_mesher.h"
-#include "Eigen/Dense"
+#include <ChiMath/chi_math.h>
```
mhanus:
```c++
#include "ChiMath/chi_math.h"
```

CHI_TECH/ChiMath/matrix_operations.cc
```diff
@@ -0,0 +1,512 @@
...
+
+  return M;
+}
+
+
+//######################################################### Power iteration
+/** Performs power iteration to obtain the fundamental eigen mode. The
+ * eigen-value of the fundamental mode is return whilst the eigen-vector
+ * is return via reference.*/
+double ChiMath::Fundamental_EigValVec_PowerIteration(
```
mhanus: The name is too complicated IMHO - I think `PowerIteration` would be enough.

CHI_TECH/ChiPhysics/PhysicsMaterial/transportxsections/prop_transportxs_03_ecollapse.cc
```diff
@@ -77,37 +78,38 @@ void chi_physics::TransportCrossSections::
...
+  MatDbl C    = math.MatMul(Ainv,B);
+  VecDbl E(G,1.0);
+  VecDbl E_new(G,0.0);
 
-  Eigen::MatrixXd Ainv = A.inverse();
-  Eigen::MatrixXd C = Ainv*B;
-  Eigen::VectorXd E(G); E.setConstant(1.0);
-  Eigen::VectorXd E_new(G);
 
   //============================================= Perform power iteration
```
mhanus, line 85: Could we use here the power iteration routine from ChiMath?

mhanus (changes requested):

CHI_TECH/ChiMath/vector_operations.cc
```diff
@@ -0,0 +1,134 @@
...
+ * \f]
+ * */
+double ChiMath::Dot(const VecDbl &x, const VecDbl &y)
+{
+  // Error Checks
+  assert(x.size() > 0);
+  assert(y.size() > 0);
+  assert(x.size() == y.size());
+  // Local Variables
+  unsigned int n = x.size();
```
wdhawkins: This should be size_t.  Several places in this commit, unsigned int is used with .size().  We should be using size_t instead.

wdhawkins:
I'd like to have a discussion about the design of this class.  In particular, I don't see the need for all of these routines to be encapsulated in a class.  I personally would think the chi_math namespace containing individual routines (and typedefs) would be a better design.

CHI_TECH/ChiPhysics/PhysicsMaterial/transportxsections/prop_transportxs_03_ecollapse.cc
Naktakala: Done. Good catch actually I thought this was already done.

CHI_TECH/ChiMath/matrix_operations.cc
Naktakala: Done.

CHI_TECH/ChiMesh/SurfaceMesher/Delaunay/delaunay_patch_00_utils.cc
Naktakala: Done

CHI_TECH/ChiMath/matrix_operations.cc
Naktakala: Done, although now I needed another fabs.

CHI_TECH/ChiMath/matrix_operations.cc
Naktakala: Done

CHI_TECH/ChiMath/matrix_operations.cc
Naktakala: Done

CHI_TECH/ChiMath/vector_operations.cc
Naktakala: Done. Replaced many references to .size() as unsigned int to size_t

Naktakala:
> I'd like to have a discussion about the design of this class. In particular, I don't see the need for all of these routines to be encapsulated in a class. I personally would think the chi_math namespace containing individual routines (and typedefs) would be a better design.

Done. I moved all matrix and vector operations to chi_math as discussed.

mhanus (approved):

wdhawkins (approved):

PR #24, opened 2019-11-03
Minor bug fix on 1D shape function [see description]

Naktakala:
Field Function interpolators use the actual xyz evaluation of the shape function to compute things like lineouts. Under certain circumstances I've noticed that the lineouts looked weird for 1D simulations and tracked it down to this bug. It luckily only affects 1D Slab geometries.


ragusa (approved):
looks fine but seems convoluted,

mhanus (approved):
I'd prefer to see `v01.Dot(xyz_ref) / (v01.Norm()*h);`, but otherwise 👍

PR #64, opened 2020-07-14
added working hypre repo to configure script for petsc configuration

jthano:


jthano:
I probably should have submitted an issue first to find out if anyone else had this problem with the dependencies

Naktakala (approved):

PR #70, opened 2020-08-12
Several changes to code base to make it easier to subclass the LBS so…

jthano:
…lver

### DIFF
--- a/Modules/LinearBoltzmanSolver/GroupSet/lbs_groupset.h
+++ b/Modules/LinearBoltzmanSolver/GroupSet/lbs_groupset.h
@@ -12,6 +12,8 @@
 
 #include <ChiPhysics/chi_physics_namespace.h>
 
+#include <functional>
+
 namespace LinearBoltzman
 {
   enum class AngleAggregationType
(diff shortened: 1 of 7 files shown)


jthano:
The changes show that a new line was added to my modified files. Do the files in the code base not have a trailing new line to end the file? Maybe this is just an issue cause by my eclipse setup

Modules/LinearBoltzmanSolver/GroupSet/lbs_groupset.h
```diff
@@ -72,6 +74,17 @@ class LBSGroupset
 
   double                                       latest_convergence_metric;
 
+  /**Convenient typdef for the moment call back function. See moment_callbacks*/
+  typedef std::function<void(int, int, int, double)> MomentCallbackF;
```
Naktakala: The only part I am missing here is what the moment arguments are. We have two ways to address it. We could either include them into the typedef:

`typedef std::function<void(int dof, int moment, int angle_num, double psi)> MomentCallbackF;`

or we could include it in its doc-string.
`/**Convenient typdef for the moment call back function. See moment_callbacks.
Arguments are:
int dof_mapping Local DOF-address with moment and group folded in.
int moment         The moment number.
int angle_num    The reference angle number in the angular quadrature.
double psi          Angular flux for the given DOF.
*/`

In any case, we will make improvements later so just choose one.

Naktakala (changes requested):
Just a commenting request. Just cause I can. Then approve

Modules/LinearBoltzmanSolver/GroupSet/lbs_groupset.h
jthano: I went ahead and updated the comment and added names to the argument list

PR #75, opened 2020-10-19
Cleanup and refactoring of PWL sweepchunk and associated files.

wdhawkins:


Modules/LinearBoltzmannSolver/SweepChunks/lbs_sweepchunk_pwl.h
```diff
@@ -170,258 +123,211 @@ class LBSSweepChunkPWL : public chi_mesh::sweep_management::SweepChunk
...
+            face_incident_flags[f] = true;
+            bool local = transport_view->face_local[f];
+            bool boundary = grid_view->IsCellBndry(face.neighbor);
+            int num_face_indices = face.vertex_ids.size();
 
-          if (mu < 0.0) //UPWIND
-          {
-            //============================== Increment face counters
-            if (neighbor_is_local)
+            if (local)
```
Naktakala, line 161: Here you split the face loop from inner branches to outer branches. I checked the performance implications on my laptop and on a xeon machine, there was no change in performance. This brings the question of why the splitting?

I prefer not having 3 different loops over face nodes in order to simplify student's understanding, as well as to reduce the number of lines of codes.

What was the particular reason for changing this?

Naktakala (approved):
It all checks out

PR #78, opened 2020-11-22
Interface fixes

jthano:
This PR includes the following items:

1) In Modules/LinearBoltzmannSolver/lua/lbs_setproperty.cc, there are a number of places where a dynamic cast is performed, but the wrong pointer is checked for validity in the subsequent if statement
2) In class LBSGroupset, I changed the function MomentCallbackF. I added two additional arguments which are a cell_id and also another dof_index. Before there was a dof index that gave the transport_view dof, but if you had your own indexing, you need information about the fe_view dof so that's what I added. Also, knowing what cell the solution is for is useful and that is given by cell_id
3) I moved sweep_orderings from LinearBoltzmann::Solver to LBSGroupset and changed it from a vector of raw pointer to unique_ptr. Because of this, I also needed to go through and change places where pointers were passed to references since you don't want to be passing unique_ptrs around. This should better reflect pointer ownership and the fact that the arguments where pointers were being passed were not optional arguments and using a reference better reflects the fact that they are required information.

### DIFF
--- a/Modules/LinearBoltzmannSolver/GroupSet/lbs_groupset.h
+++ b/Modules/LinearBoltzmannSolver/GroupSet/lbs_groupset.h
@@ -78,18 +78,19 @@ public:
   /**
    * Convenient typdef for the moment call back function. See moment_callbacks.
    *  Arguments are:
-   *  int dof_mapping Local DOF-address with moment and group folded in.
-   *  int moment The moment number.
-   *  int angle_num The reference angle number in the angular quadrature
-   *  double psi Angular flux for the given dof and angle_num.
+   *  int cell_id, the value cell->local_id for the current cell
+   *  int dof_mapping, local DOF-address with moment and group folded in.
+   *  int dof_index, dof for the solution psi from ((CellFEView*)grid_fe_view->MapFeViewL(cell->local_id))->dofs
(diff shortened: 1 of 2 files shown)


jthano:
The important of item number 3 was moving sweep_orderings from LinearBoltzmann::Solver to LBSGroupset. So if changing sweep_orderings from a vector of raw pointers to unique pointers introduced too many changes, I can go back and make it a vector of raw pointers again

wdhawkins:
> The important of item number 3 was moving sweep_orderings from LinearBoltzmann::Solver to LBSGroupset. So if changing sweep_orderings from a vector of raw pointers to unique pointers introduced too many changes, I can go back and make it a vector of raw pointers again

I would vote to keep this change. We should be using managed/smart pointers throughout the code. If there is concern about the number of changes, passing the stored pointer to objects and methods is perfectly acceptable and still maintains the concept of ownership. Granted, that concept might be a bit muddy right now since we do make extensive use of raw pointers, but it will become more clear as we refactor the code to use managed pointers.

ragusa:
Good point. I wanted to specifically talk about this during this morning’s meeting.

jthano:
I removed the smart pointer stuff from my PR. This smaller PR is ready for review

PR #112, opened 2021-04-08
LBS - obtain number of moments directly from angular quadrature

dominiccaron:


ChiModules/LinearBoltzmannSolver/lbs_01ab_compute_nummoments.cc
```diff
@@ -1,30 +1,15 @@
...
-
-extern ChiLog& chi_log;
-
 //###################################################################
 /** Computes the number of moments for the given mesher types*/
 void LinearBoltzmann::Solver::ComputeNumberOfMoments()
 {
-  int L = options.scattering_order;
+  num_moments =
+    group_sets.front().quadrature->GetMomentToHarmonicsIndexMap().size();
```
Naktakala: I've been thinking about this one and here is my concern. A user can screw up and provide two different custom quadratures (like what Grissom is doing right now) which could result in two different number of moments. If the first groupset has `num_moments` as some low number (say 15) and then a subsequent groupset had a higher amount for `num_moments` then there would be corruption when addressing vector addresses. In the opposite sense results could be garbage.

I would like to see a loop after this line over all groupsets. In the loop we check to see if all the groupsets have the same `MomentToHarmonicsIndexMap`.

Example:
```
void LinearBoltzmann::Solver::ComputeNumberOfMoments()
{
  auto& gs0_quadrature = group_sets.front().quadrature;

  num_moments = (int)gs0_quadrature->GetMomentToHarmonicsIndexMap().size();

  auto& m_to_ell_em_map = gs0_quadrature->GetMomentToHarmonicsIndexMap();
  for (auto& group_set : group_sets)
    if (group_set.quadrature->GetMomentToHarmonicsIndexMap() != m_to_ell_em_map)
      throw std::logic_error("LinearBoltzmann::Solver::ComputeNumberOfMoments : "
                             "Moment-to-Harmonics mapping differs between "
                             "groupsets and is not allowed.");

  if (num_moments == 0)
    throw std::logic_error("LinearBoltzmann::Solver::ComputeNumberOfMoments : "
                           "unable to infer number of moments from angular "
                           "quadrature.");
}
```

The above code won't compile unless we add the code below to the `HarmonicIndices` struct:

```
bool operator==(const HarmonicIndices& other) const
    {
      return (ell == other.ell and m == other.m);
    }
```

Naktakala (changes requested):

ChiModules/LinearBoltzmannSolver/lbs_01ab_compute_nummoments.cc
dominiccaron: Correct; commit 31bd5980e9c60655de992ef0f23149fa38d22c2b operates on the same principle as the previous implementation of the method ComputeNumberOfMoments, which is insufficient to catch the scenario that you mention. This modification renders the method more robust.

Naktakala (approved):

PR #117, opened 2021-04-28
Changes to precursor handling

zachhardy:
This PR changes the way precursors are handled to allow for multi-material problems and adds an initialization routine for the precursor concentrations after the execution of the solver in the event that data is desired.

TransportCrossSections now has an additional member variable, precursor_map. This is a vector of integers meant to hold the global IDs of the precursors that live on a material. In a problem with 2 materials each with 6 precursor groups, 12 precursors must be tracked. In this example, the global IDs 0-5 are assigned to the first material and the global IDs 6-11 are assigned to the second. While the material properties are still accessed via the local ID of the precursor, the precursor_map object provides a mapping to the global precursor vector. These global IDs are distributed to the TransportCrossSections objects during the initialization routine of the solver.


CMakeLists.txt
```diff
@@ -124,7 +124,7 @@ string(REPLACE "-O2" "-O3" CMAKE_CXX_FLAGS_RELEASE ${CMAKE_CXX_FLAGS_RELEASE})
 
 #================================================ Define source directories
 set(SOURCES "${CHI_TECH_DIR}/ChiTech/chi_runtime.cc"
-            "${CHI_TECH_DIR}/ChiTech/LuaTest/lua_test.cc")
+            "${CHI_TECH_DIR}/ChiTech/LuaTest/lua_test.cc" ChiModules/KEigenvalueSolver/initialize_precursors.cc)
```
Naktakala: This change is redundant. There is a `CMakeLists.txt` file in that directory that already covers it. I checked it so just revert this change.

ChiModules/KEigenvalueSolver/initialize_k_solver.cc
```diff
@@ -8,16 +13,40 @@ void KEigenvalue::Solver::InitializeKSolver()
...
+
+    // -----Initialize precursor unknown manager and vector
+    if (num_precursors > 0) {
+      auto pwl = std::static_pointer_cast<SpatialDiscretization_PWLD>(discretization);
+      Nj_unk_man.AddUnknown(chi_math::UnknownType::VECTOR_N, num_precursors);
+
+      int local_Nj_dof_count = pwl->GetNumLocalDOFs(Nj_unk_man);
+      Nj_new_local.resize(local_Nj_dof_count, 0.0);
+    }
+  }
```
Naktakala, line 51: Clever.

Naktakala (changes requested):

PR #119, opened 2021-04-30
Improvements to VTU unpartitioned mesh reader

dominiccaron:
Improvements to VTU unpartitioned mesh reader:
- ability to read one-, two- and three-dimensional meshes;
- ability to read material identifier and boundary identifier.

In reading a mesh of arbitrary dimension, the present algorithm involves two complete passes through the mesh: (1) to determine the maximum dimension of the cells of the mesh (assuming unordered cells) and (2) to reconstruct all cells encountered in the mesh file. The latter can be rendered more efficient if cells of dimension less than DIM-1 are bypassed, with DIM the maximum dimension of cells determined in the first pass. However, that optimisation may be best only after abandoning the present approach of counting cells by type that are encountered in the VTU file.

The reading of material identifiers and boundary identifiers assumes that the VTU mesh file may contain a vtkCellData field of arbitrary name and arbitrary type in which cell identifiers are stored. If the user provides the name of this field on input, the VTU reader reads the cell identifier from this field after having read the cell itself. Note that the GetTuple method of the class vtkDataArray is employed to extract the data from the field containing the cell identifier. This approach simplifies implementation but will produce incorrect results if the vtkCellData field of the VTU file contains non-integral values.

### DIFF
--- a/ChiTech/ChiMesh/UnpartitionedMesh/lua/create.cc
+++ b/ChiTech/ChiMesh/UnpartitionedMesh/lua/create.cc
@@ -11,6 +11,8 @@
 /**Creates an unpartitioned mesh from VTK Unstructured mesh files.
 
 \param file_name char Filename of the .vtu file.
+\param field char Name of the cell data field from which to read
+                  material and boundary identifiers (optional).
 
 \ingroup LuaUnpartitionedMesh
 
@@ -39,14 +41,18 @@ int chiUnpartitionedMeshFromVTU(lua_State* L)
 {
   const char func_name[] = "chiUnpartitionedMeshFromVTU";
   int num_args = lua_gettop(L);
-  if (num_args != 1)
+  if (num_args < 1)
     LuaPostArgAmountError(func_name,1,num_args);
 
   const char* temp = lua_tostring(L,1);
+  const char* field = "";
+  if (num_args >= 2) field = lua_tostring(L,2);
   auto new_object = new chi_mesh::UnpartitionedMesh;
 
   chi_mesh::UnpartitionedMesh::Options options;
   options.file_name = std::string(temp);
+  options.material_id_fieldname = field;
+  options.boundary_id_fieldname = field;
 
   new_object->ReadFromVTU(options);
 
--- a/ChiTech/ChiMesh/UnpartitionedMesh/lua/create.cc
+++ b/ChiTech/ChiMesh/UnpartitionedMesh/lua/create.cc
@@ -44,6 +44,9 @@ int chiUnpartitionedMeshFromVTU(lua_State* L)
   if (num_args < 1)
     LuaPostArgAmountError(func_name,1,num_args);
 
+  LuaCheckNilValue(func_name,L,1);
+  if (num_args >= 2) LuaCheckNilValue(func_name,L,2);
+
   const char* temp = lua_tostring(L,1);
   const char* field = "";
   if (num_args >= 2) field = lua_tostring(L,2);
@@ -100,6 +103,9 @@ int chiUnpartitionedMeshFromEnsightGold(lua_State* L)
   if (num_args <1)
     LuaPostArgAmountError(func_name,1,num_args);
 
+  LuaCheckNilValue(func_name,L,1);
+  if (num_args >= 2) LuaCheckNilValue(func_name,L,2);
+
   const char* temp = lua_tostring(L,1);
   double scale = 1.0;
   if (num_args >= 2) scale = lua_tonumber(L,2);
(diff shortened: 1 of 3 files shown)


ChiTech/ChiMesh/UnpartitionedMesh/lua/create.cc
```diff
@@ -39,14 +41,18 @@ int chiUnpartitionedMeshFromVTU(lua_State* L)
 {
   const char func_name[] = "chiUnpartitionedMeshFromVTU";
   int num_args = lua_gettop(L);
-  if (num_args != 1)
+  if (num_args < 1)
     LuaPostArgAmountError(func_name,1,num_args);
```
Naktakala, line 46: Please add two lines here. 

```
LuaCheckNilValue(func_name,L,1);
if (num_args >=2) LuaCheckNilValue(func_name,L,2);
```

ChiTech/ChiMesh/UnpartitionedMesh/lua/create.cc
```diff
@@ -39,14 +41,18 @@ int chiUnpartitionedMeshFromVTU(lua_State* L)
...
 
   const char* temp = lua_tostring(L,1);
+  const char* field = "";
+  if (num_args >= 2) field = lua_tostring(L,2);
```
ragusa, line 52: should this be == rather than >= ?

ChiTech/ChiMesh/UnpartitionedMesh/unpartmesh_01_readfromvtu.cc
```diff
@@ -255,6 +259,63 @@ CreateCellFromVTKTriangle(vtkCell *vtk_cell)
...
+//###################################################################
+/**Creates a raw point cell from a vtk-vertex.*/
+chi_mesh::UnpartitionedMesh::LightWeightCell* chi_mesh::UnpartitionedMesh::
+  CreateCellFromVTKVertex(vtkCell *vtk_cell)
+{
+  auto point_cell  = new LightWeightCell(chi_mesh::CellType::GHOST);
+
+  auto vtk_vertex  = vtkVertex::SafeDownCast(vtk_cell);
+  auto num_cpoints = vtk_vertex->GetNumberOfPoints();
+  auto num_cfaces  = num_cpoints;
```
Naktakala: `num_cfaces` is not used. Please remove.

ragusa (approved):
I am a bit unsure about the GetTuple comments. Is this safe?

ChiTech/ChiMesh/UnpartitionedMesh/unpartmesh_01_readfromvtu.cc
```diff
@@ -300,25 +361,94 @@ void chi_mesh::UnpartitionedMesh::
...
-  for (int c=0; c<total_cell_count; ++c)
-    if (CellIs3D(ugrid->GetCell(c)->GetCellType()))
+    chi_log.Log(LOG_ALLERROR)
+      << "The VTU reader expects material identifiers and boundary identifiers "
+      << "to be defined in the same field.";
+    std::exit(EXIT_FAILURE);
+  }
+
+  vtkDataArray* cell_id_array_ptr = nullptr;
+  if (options.material_id_fieldname.size() == 0)
```
Naktakala: Clang-tidy indicates to use `options.material_id_fieldname.empty()` instead of `size()==0`.

Naktakala (changes requested):

ChiTech/ChiMesh/UnpartitionedMesh/lua/create.cc
Naktakala: That's fine cause extra arguments should be ignored, not alter behavior. If a third argument is added by accident the >= will still work but not ==

ChiTech/ChiMesh/UnpartitionedMesh/lua/create.cc
Naktakala: These are checks that are actually needed across-the-board. Let's try and add them wherever we can.

dominiccaron: You must have anticipated my response. Included the verification in all four methods defined in the file in question.

ChiTech/ChiMesh/UnpartitionedMesh/lua/create.cc
dominiccaron: At the moment, there exists another method defined in the same file that accepts an optional argument and it employs the same logic as above; the line identified has been left as-is for consistency.

ChiTech/ChiMesh/UnpartitionedMesh/unpartmesh_01_readfromvtu.cc
dominiccaron: Definition in libc++ (https://github.com/llvm/llvm-project/blob/main/libcxx/include/string#L984):
```
    bool empty() const _NOEXCEPT {return size() == 0;}
```
Likewise, definition in libstdc++ is equivalent.

Naktakala (approved):

PR #122, opened 2021-05-03
Intel compiler fixes

Naktakala:
The intel c++ compiler can be a little more pedantic about certain things like nested namespaces. These changes mostly fixed this.

I also threw in some minor cosmetic things while I was checking related files.


ragusa (approved):
I see nothing wrong here. A little unsure about the changes about namespaces, but this strikes me as an implementation strategy change.

dominiccaron:
The behaviour is justified: the ability to define a nested namespace using the syntax ```namespace A::B { ... }``` as an alternative to ```namespace A { namespace B { ... } }``` is available in C++17 onward. Otherwise, everything looks good.

dominiccaron (approved):

wdhawkins (approved):

PR #123, opened 2021-05-05
Refactor PWL cell mappings and PWL spatial discretisations

dominiccaron:
Refactor of the PWL cell mappings and of the PWL spatial discretisations, including:
- modification of the one-dimensional Gaussian quadratures to manage self-consistently the interval on which they are defined;
- transition to a line quadrature object (Gauss-Legendre quadrature defined on [0;1]) for spatial integration in one-dimension to be consistent with the quadratures employed for multi-dimensional spatial integration;
- corrections to the quadrature point data interface and to the initialisation of quadrature point data for the various cell mappings;
- implementation of a generic method to compute cell matrices from quadrature point data and transition to this method for all cell types except slab;
- removal of the auxiliary quadratures in the PWL cell mappings and in the PWL spatial discretisations;
- cleaning as a consequence of the aforementioned points.


Naktakala (approved):

ChiTech/ChiMath/SpatialDiscretization/FiniteElement/PiecewiseLinear/pwl_00_constrdestr.cc
```diff
@@ -35,6 +30,36 @@ SpatialDiscretization_PWLD::
                 << " Communicating partition neighbors.";
   ref_grid->CommunicatePartitionNeighborCells(neighbor_cells);
 
+  if (setup_flags == chi_math::finite_element::COMPUTE_UNIT_INTEGRALS)
+  {
+    const auto qorder_min = (int)chi_math::QuadratureOrder::SECOND;
```
wdhawkins: We should use proper C++ style casts.

wdhawkins (changes requested):
We should be using proper C++ style casts.

ChiTech/ChiMath/SpatialDiscretization/FiniteElement/PiecewiseLinear/pwl_00_constrdestr.cc
dominiccaron: Agreed; in this case, likely I was attempting to match the convention employed elsewhere in the class. I have revised these instances.

ragusa (approved):

ChiTech/ChiMath/SpatialDiscretization/FiniteElement/PiecewiseLinear/pwl_00_constrdestr.cc
Naktakala: I am a bit lost here. What was the old code, and what did it need to be changed to?

dominiccaron: See 09112e3cf74d2a26e2b58d72a186a806fd4f9113.

Naktakala: Ahhh, I meant for the change to be spelled out here. Just for easier referencing and for others to learn from, I am posting it here again:

The casting using `(int)` should be replaced with `static_cast<int>`.

A valid question is: Why? Short answer: it is good practice. The more general convention of when c-style casting can be acceptable needs a more intricate answer which we should pick up in a discussion on the boards.

Naktakala (approved):

wdhawkins (approved):

PR #125, opened 2021-05-16
Spatial Discretization now has GetCellNumNodes

Naktakala:
In general the type of spatial discretization will dictate the number of nodes placed on a cell. For example, FV will place one node per cell and PWL or PWLD will place one node per cell-vertex. The only way to get this information for a single cell is to get its mapping or unit-integrals which is an expensive route. The method `GetCellNumNodes` has now been added to the base `SpatialDiscretization` class to general this notion and to remove the need to cast a discretization to its child so that the necessary information can be obtained during the sizing of vectors.


ChiTech/ChiMath/SpatialDiscretization/spatial_discretization.h
```diff
@@ -71,6 +71,9 @@ class SpatialDiscretization
   size_t GetNumGlobalDOFs(chi_math::UnknownManager& unknown_manager)
                           {return 0;}
 
+  virtual
+  size_t GetCellNumNodes(const chi_mesh::Cell& cell) = 0;
```
dominiccaron: I propose that this method be marked ```const```.

dominiccaron (approved):

ChiTech/ChiMath/SpatialDiscretization/spatial_discretization.h
Naktakala: Yes! Totally forgot.

Naktakala: Done.

wdhawkins (commented):
I don't know about the NumNodes terminology.  To me, it implies a geometric connection... which may not always be accurate.  Why not NumDOF or something similar?

ragusa:
If it used to describes vertices, then use node in the name. If it is meant to describe the DOFs, the size of the associated linear system, then use DOF in the name.

I don’t know if you recall, but with Mike Hackemack, we did quadratic Pwld on arbitrary polygons. Mike then improved upon his PhD work and maybe quadratic on extruded arbitrary polygons. So it is possible/imaginable that, one day, we put quadratic pwld in chi-tech.



—jean ragusa


http://multiphysics.engr.tamu.edu
http://nustem.engr.tamu.edu
-.-.-
iPhone, iTypo, iApologize.

On May 17, 2021, at 10:05 AM, wdhawkins ***@***.***> wrote:

﻿

@wdhawkins commented on this pull request.

I don't know about the NumNodes terminology. To me, it implies a geometric connection... which may not always be accurate. Why not NumDOF or something similar?

Naktakala:
Very important comment Daryl, at least to me. Passionate topic for me. There is a lot of confusion with the connection between vertices, nodes and DOFs. Here is my hierarchy of definitions that is thus far consistent across a couple of textbooks on FE:

- `Vertices` are directly connected to cell-geometry. Right now we only support cells that have exterior vertices (which needs to change if we wish to support cells with higher order geometrical representations). _The geometrical representation allows us to compute things like volumes, areas, and Jacobians._ Slabs right now are easy because they only have 2 exterior vertices. Triangles, Quads and Polygons have their vertices listed in right-hand rule notation, and their faces are listed as edges with vertex order consistent with a right-hand rule triangle formed with the cell centroid. 3D elements have unique orderings if they are standard elements but their faces follow 2D rules. With higher-order elements we will need to be a lot more careful to get the conventions right (right-hand rule kind of breaks with interior vertices).

- `Nodes` are locations to which we associate one or more degrees-of-freedom (DOFs). These nodes may be the same as the vertices but they can also be less/more in number and in different locations. Finite Volume discretizations have only a single node regardless of what type of cell it is. In the case of a Finite Element spatial discretization, where we represent a field variable using the same interpolation function as the underlying geometry interpolation function, we call the type of FE-elements `Isoparametric-elements`. _But this is not always the case_. If we want to apply a higher-order FE discretization we require more support points for the interpolation functions and therefore have to add more nodes that do not coincide with the vertices, hence this case calls for `Superparametric-elements`.

- `Degrees-Of-Freedom` or Nodal-variables are the ultimate field-variable unknowns (i.e. flux_m0_g0, flux_m0_g1, etc.). DOFs are stacked onto nodes. Once the nodal-structure of a discretization is known we can play around with different DOF-structures without adding a lot more work when we need to map addresses and stuff. For instance a single PWLD discretization can serve multiple DOF structures. If we use `MapDOF` without supplying a `UnknownManager` the discretization assumes a single DOF per node, but if we do supply an `UnknownManager` the discretization can use the underlying nodal structure to map specific variable addresses without having to initialize parallel addresses differently. The `UnknownManager` essentially allows you flexibility with Vectors and Matrices both in a Nodal-ordering and a Block-ordering.

We need to later add `NumDOFs` too but that requires an `UnknownManager` similar to how deal.ii requires a `DOFHandler`. In our transport context we need the base `NumNodes` because we are hopscotch-exchanging DOF structure when we run different groupsets.

Very interested in comments on this subject.

Naktakala:
> I don't know about the NumNodes terminology. To me, it implies a geometric connection... which may not always be accurate. Why not NumDOF or something similar?

Did my above comment clarify your question Daryl?

wdhawkins:
> > I don't know about the NumNodes terminology. To me, it implies a geometric connection... which may not always be accurate. Why not NumDOF or something similar?
> 
> Did my above comment clarify your question Daryl?

It did.  I very much like the descriptions you provided.

wdhawkins (approved):

PR #130, opened 2021-05-20
Bug fix for combined cross sections

Naktakala:
Small bug fix when no fission cross-sections are present and user attempts to combine cross-sections.

### DIFF
--- a/ChiTech/ChiPhysics/PhysicsMaterial/transportxsections/prop_transportxs_00_general.cc
+++ b/ChiTech/ChiPhysics/PhysicsMaterial/transportxsections/prop_transportxs_00_general.cc
@@ -133,8 +133,6 @@ void chi_physics::TransportCrossSections::
       this->is_fissile = true;
       Nf_total += combo.second;
     }
-    if (Nf_total < 1.0e-28)
-      Nf_total = 1.0; //Avoids divide by 0 when non-fissile
 
     //============================ Check number of groups
     if (cross_secs.size() == 1)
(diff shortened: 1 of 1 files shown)


ragusa:
to combine cross-sections -->  to combine cross sections.  (No hyphen).

ragusa (approved):

ragusa (approved):
So, the atom density is in units of per barn-cm, right? Just making sure I understand the magnitude chosen for this test.

dominiccaron (commented):
Would it not be better to make the reassignment after the completion of the ```for (auto combo : combinations)``` that begins on line 114, but before the statement ```double ff_i = N_i/Nf_total;``` on line 211? As ```Nf_total``` is accumulated within the for-loop identified above, if the direct assignment of ```1.0``` occurs within the for-loop, it can produce an incorrect value of the summation in various scenarios when it is not truly zero (or less than the imposed threshold).

Naktakala:
> to combine cross-sections --> to combine cross sections. (No hyphen).

What was the convention again? cross-section and cross sections? Singular vs Plural?

Naktakala:
> Would it not be better to make the reassignment after the completion of the `for (auto combo : combinations)` that begins on line 114, but before the statement `double ff_i = N_i/Nf_total;` on line 211? As `Nf_total` is accumulated within the for-loop identified above, if the direct assignment of `1.0` occurs within the for-loop, it can produce an incorrect value of the summation in various scenarios when it is not truly zero (or less than the imposed threshold).

You are correct. Watch how code-review just saved another bug.

dominiccaron (approved):

wdhawkins (approved):

Naktakala:
> So, the atom density is in units of per barn-cm, right? Just making sure I understand the magnitude chosen for this test.

This class-unit is essentially dimensionless but LBS uses it as per barn-cm.

ragusa:
cross-section library: hyphen because XS acts as an adjective in the compound noun

total cross section: no hyphe, XS acts as a noun

PR #131, opened 2021-05-20
Source flags and src scope

Naktakala:
This PR changes quite a bit of files. Please read the highlights below before reviewing to better comprehend the impact of the minor items.

# Highlights
- `SweepChunk*` changed to `std::shared_ptr<SweepChunk>` in `SetSweepChunk` routine, and references are now passed where applicable, instead of the raw pointer.
- `SourceFlags` is now an enum.
- Some renaming to keep consistency with nodes-vs-DOFs
- Ad-hoc renamings (not a lot)
- Somewhat of a cleanup of the source routines (both `SetSource` and `SetKSource`.
- Reorganization of GMRES
- `KSPDataContext` and `LBSMatrixAction_Ax` has now been appended to the `LinearBoltzmann` namespace
- Replaced all the pointers in `KSPDataContext` with references and added a constructor
- Removed an old unused `KSPMonitorNPT` routine

# Details regarding source flags
- Adding the material-source, the scattering-source and the fission-source to the `q_moments`-vector is an ability included in the `SetSource` routine. The first vexing thing was that we had this strange yes-no logic for only two of these items where we had an `apply_mat_src`-flag and a `suppress_phi_old`-flag, so if you wanted both the material-source and the scattering-source you had to say `SetSource(groupset, true, false)`. The second thing was that fission was operated via the `suppress_phi_old`-flag. These flags are now all consolidated into an integer bit-set which will support further future options.

### Old code example GMRES
```
//Computing b
SetSource(groupset, SourceFlags::USE_MATERIAL_SOURCE, SourceFlags::SUPPRESS_PHI_OLD);

//Inside matrix action
SetSource(groupset, LinearBoltzmann::SourceFlags::USE_DLINV_SOURCE, false);
```
### New code example GMRES
```
//Computing b
SourceFlags rhs_src_scope = APPLY_MATERIAL_SRC;
SetSource(groupset, rhs_src_scope);

//Inside matrix action
SourceFlags lhs_src_scope = APPLY_SCATTER_SOURCE | APPLY_FISSION_SOURCE;
SetSource(groupset, lhs_src_scope);
```

GMRES also now gets called with the flags for `lhs_src_scope` and `rhs_src_scope` adding a lot of versatility to any call to GMRES. Examples:
- One can do GMRES as usual `rhs_src_scope = APPLY_MATERIAL_SRC` and `lhs_src_scope = APPLY_SCATTER_SOURCE | APPLY_FISSION_SOURCE`
- One can do GMRES to purely converge cyclic angular fluxes `rhs_src_scope = APPLY_MATERIAL_SRC | APPLY_SCATTER_SOURCE | APPLY_FISSION_SOURCE` and `lhs_src_scope = NO_FLAGS_SET`

One can also theoretically call `LBSMatrixAction_Ax`-standalone although this is not ready yet since it get the context from a matrix shell. But this will soon be split into an operator-action section and the matrix action.

### DIFF
--- a/ChiModules/KEigenvalueSolver/IterativeMethods/power_iteration.cc
+++ b/ChiModules/KEigenvalueSolver/IterativeMethods/power_iteration.cc
@@ -33,7 +33,7 @@ void KEigenvalue::Solver::PowerIteration(LBSGroupset& groupset)
     << "the Power Method.\n\n";
 
   // ----- Setting up required sweep chunks
-  SweepChunk* sweep_chunk = SetSweepChunk(groupset);
+  auto sweep_chunk = SetSweepChunk(groupset);
 
   // ----- Set sweep scheduler
   MainSweepScheduler SweepScheduler(SchedulingAlgorithm::DEPTH_OF_GRAPH,
@@ -70,10 +70,12 @@ void KEigenvalue::Solver::PowerIteration(LBSGroupset& groupset)
     for (int si_nit=0; si_nit<groupset.max_iterations; si_nit++)
     {
       // ----- Set source and sweep
-      SetKSource(groupset, SourceFlags::USE_MATERIAL_SOURCE);
+      SetKSource(groupset, APPLY_MATERIAL_SOURCE |
+                           APPLY_SCATTER_SOURCE |
+                           APPLY_FISSION_SOURCE);
       groupset.angle_agg.ZeroOutgoingDelayedPsi();
       phi_new_local.assign(phi_new_local.size(),0.0);
-      SweepScheduler.Sweep(sweep_chunk);
+      SweepScheduler.Sweep(*sweep_chunk);
 
       // ----- Compute convergence parameters
       pw_change = ComputePiecewiseChange(groupset);
@@ -156,8 +158,8 @@ void KEigenvalue::Solver::PowerIteration(LBSGroupset& groupset)
     chi_log.ProcessEvent(source_event_tag,
                          ChiLog::EventOperation::AVERAGE_DURATION);
   size_t num_angles = groupset.quadrature->abscissae.size();
-  long int num_unknowns = (long int)glob_dof_count*
-                          (long int)num_angles*
+  long int num_unknowns = (long int)glob_node_count *
+                          (long int)num_angles *
                           (long int)groupset.groups.size();
   chi_log.Log(LOG_0)
     << "\n";
--- a/ChiModules/KEigenvalueSolver/IterativeMethods/power_iteration.cc
+++ b/ChiModules/KEigenvalueSolver/IterativeMethods/power_iteration.cc
@@ -48,12 +48,10 @@ void KEigenvalue::Solver::PowerIteration(LBSGroupset& groupset)
                                          phi_old_local.data()); 
 
   // ----- Start outer k iterations
-  double F_new = 0.0;
-  double F_prev = 1.0;
+  double F_new = 0.0;        //Production source new (0.0 is not used)
+  double F_prev = 1.0;       //Production source prev
   double k_eff_prev = k_eff;
-  double k_eff_change = 0.0;
-  double reactivity = 0.0;
-  int nit = 0;
+  int nit = 0;               //number of iterations
   bool k_converged = false;
 
   while (nit < options.max_iterations)
@@ -63,9 +61,7 @@ void KEigenvalue::Solver::PowerIteration(LBSGroupset& groupset)
 
 
     // ----- Start inner source iterations
-    double pw_change = 0.0;
     double pw_change_prev = 1.0;
-    double rho = 0.0;
     bool si_converged = false;
     for (int si_nit=0; si_nit<groupset.max_iterations; si_nit++)
     {
@@ -78,10 +74,10 @@ void KEigenvalue::Solver::PowerIteration(LBSGroupset& groupset)
       SweepScheduler.Sweep(*sweep_chunk);
 
       // ----- Compute convergence parameters
-      pw_change = ComputePiecewiseChange(groupset);
+      double pw_change = ComputePiecewiseChange(groupset);
       DisAssembleVectorLocalToLocal(groupset,phi_new_local.data(),
                                              phi_old_local.data());
-      rho = sqrt(pw_change/pw_change_prev);
+      double rho = sqrt(pw_change/pw_change_prev);
       pw_change_prev = pw_change;
       nit += 1;
 
@@ -124,10 +120,10 @@ void KEigenvalue::Solver::PowerIteration(LBSGroupset& groupset)
     // ----- Recompute eigenvalue
     F_new = ComputeProduction();
     k_eff = F_new/F_prev * k_eff;
-    reactivity = (k_eff - 1.0) / k_eff;
+    double reactivity = (k_eff - 1.0) / k_eff;
 
     // ----- Compute convergence parameters and bump values
-    k_eff_change = fabs(k_eff - k_eff_prev) / k_eff;
+    double k_eff_change = fabs(k_eff - k_eff_prev) / k_eff;
     k_eff_prev = k_eff; F_prev = F_new;
     DisAssembleVectorLocalToLocal(groupset,phi_new_local.data(),
                                            phi_prev_local.data());
@@ -158,9 +154,9 @@ void KEigenvalue::Solver::PowerIteration(LBSGroupset& groupset)
     chi_log.ProcessEvent(source_event_tag,
                          ChiLog::EventOperation::AVERAGE_DURATION);
   size_t num_angles = groupset.quadrature->abscissae.size();
-  long int num_unknowns = (long int)glob_node_count *
-                          (long int)num_angles *
-                          (long int)groupset.groups.size();
+  size_t num_unknowns = glob_node_count *
+                        num_angles *
+                        groupset.groups.size();
   chi_log.Log(LOG_0)
     << "\n";
   chi_log.Log(LOG_0)
@@ -174,7 +170,7 @@ void KEigenvalue::Solver::PowerIteration(LBSGroupset& groupset)
     << sweep_time;
   chi_log.Log(LOG_0)
     << "        Sweep Time/Unknown (ns):       "
-    << sweep_time*1.0e9*chi_mpi.process_count/num_unknowns;
+    << sweep_time*1.0e9*chi_mpi.process_count/static_cast<double>(num_unknowns);
   chi_log.Log(LOG_0)
     << "        Number of unknowns per sweep:  " << num_unknowns;
   chi_log.Log(LOG_0)
--- a/ChiModules/KEigenvalueSolver/IterativeOperations/set_k_source.cc
+++ b/ChiModules/KEigenvalueSolver/IterativeOperations/set_k_source.cc
@@ -56,7 +56,7 @@ void KEigenvalue::Solver::
     int cell_matid = cell.material_id;
     int xs_id = matid_to_xs_map[cell_matid];
 
-    if ((xs_id<0) || (xs_id>=material_xs.size()))
+    if ( (xs_id<0) || (xs_id>=material_xs.size()) )
     {
       chi_log.Log(LOG_ALLERROR)
       << "Cross-section lookup error\n";
@@ -85,7 +85,7 @@ void KEigenvalue::Solver::
         {
           // ----- Contribute scattering
           double inscat_g = 0.0;
-          if ((ell < xs->transfer_matrix.size()) && (apply_scatter_src) )
+          if ( (ell < xs->transfer_matrix.size()) && (apply_scatter_src) )
           {
             size_t num_transfers = xs->transfer_matrix[ell].rowI_indices[g].size();
             for (int t=0; t<num_transfers; t++)
@@ -99,7 +99,7 @@ void KEigenvalue::Solver::
 
           // ----- Contribute fission
           double fission_g = 0.0;
-          if ((ell == 0) and (apply_fission_src))
+          if ( (ell == 0) and (apply_fission_src) )
           {
             if (xs->is_fissile)
             {
@@ -116,9 +116,9 @@ void KEigenvalue::Solver::
 
           // ----- Contribute precursors
           double precursor_g = 0.0;
-          if ((ell == 0) and (options.use_precursors))
+          if ( (ell == 0) and (options.use_precursors) )
           {
-            if ((apply_mat_src) and (xs->J > 0))
+            if ( (apply_mat_src) and (xs->J > 0) )
             {
               for (int j=0; j<num_precursors; ++j)
               {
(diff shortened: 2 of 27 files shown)


ChiModules/KEigenvalueSolver/IterativeMethods/power_iteration.cc
```diff
@@ -156,8 +158,8 @@ void KEigenvalue::Solver::PowerIteration(LBSGroupset& groupset)
     chi_log.ProcessEvent(source_event_tag,
                          ChiLog::EventOperation::AVERAGE_DURATION);
   size_t num_angles = groupset.quadrature->abscissae.size();
-  long int num_unknowns = (long int)glob_dof_count*
-                          (long int)num_angles*
+  long int num_unknowns = (long int)glob_node_count *
```
ragusa: is this an example where C++ casting should be used?

ChiModules/KEigenvalueSolver/IterativeOperations/set_k_source.cc
```diff
@@ -121,14 +115,14 @@ void KEigenvalue::Solver::SetKSource(LBSGroupset& groupset,
           q_mom[g] += fission_g;
 
           // ----- Contribute precursors
-          precursor_g = 0.0;
+          double precursor_g = 0.0;
           if ((ell == 0) and (options.use_precursors))
           {
             if ((apply_mat_src) and (xs->J > 0))
```
ragusa: I guess J is the # of precursors. Could we use a better name for that variable than just J?

ChiModules/KEigenvalueSolver/IterativeOperations/set_k_source.cc
```diff
@@ -80,36 +74,36 @@ void KEigenvalue::Solver::SetKSource(LBSGroupset& groupset,
...
+        double* phi_prevp = &phi_prev_local[ir];
 
         // ----- Loop over groupset groups
         for (int g=gs_i; g<=gs_f; g++)
         {
           // ----- Contribute scattering
-          inscat_g = 0.0;
-          if ((ell < xs->transfer_matrix.size()) && (!suppress_phi_old) )
+          double inscat_g = 0.0;
+          if ((ell < xs->transfer_matrix.size()) && (apply_scatter_src) )
```
ragusa: Why is apply_scatter_src for?

ragusa: spacing and parentheses is not consistent.

ChiModules/LinearBoltzmannSolver/IterativeOperations/lbs_setsource.cc
```diff
@@ -65,54 +68,49 @@ void LinearBoltzmann::Solver::SetSource(LBSGroupset& groupset,
...
-                sigma_sm  = xs->transfer_matrix[ell].rowI_values[g][t];
+                double sigma_sm = xs->transfer_matrix[ell].rowI_values[g][t];
                 inscat_g += sigma_sm * phi_oldp[gprime];
               }
             }
           }//if moment avail
 
           //====================== Apply within-groupset scattering
-          if ((ell < xs->transfer_matrix.size()) && (!suppress_phi_old) )
+          if ((ell < xs->transfer_matrix.size()) && (apply_scatter_src) )
```
ragusa, line 105: parenthesis

ChiModules/KEigenvalueSolver/IterativeOperations/set_k_source.cc
Naktakala: I agree. This however would require a number of changes not in the scope of this PR. Would you be ok with me creating an issue for it to the code author (Zach Hardy)?

ChiModules/KEigenvalueSolver/IterativeOperations/set_k_source.cc
Naktakala: You probably mean "What is apply_scatter_src for?". It is a means to turn off certain portions of the `SetKSource` routine similar to what the `SetSource` routine does. The usefulness is not immediately apparent since we don't have any other methods other than `PowerIteration` but unless you have a strong objection against I would rather leave it untouched until we either can confirm it is not needed or find a better way to use the routine.

ChiModules/KEigenvalueSolver/IterativeMethods/power_iteration.cc
Naktakala: Yes. This code was changed as part of a refactor but I went ahead and changed the obvious c-style casts in the file. Also cleaned it up a bit on some places where clang-tidy complained.

ChiModules/KEigenvalueSolver/IterativeOperations/set_k_source.cc
Naktakala: Corrected.

ChiModules/LinearBoltzmannSolver/IterativeOperations/lbs_setsource.cc
Naktakala: Corrected

dominiccaron (approved):

ChiModules/LinearBoltzmannSolver/IterativeMethods/lbs_GMRES.cc
```diff
@@ -40,65 +44,53 @@ bool LinearBoltzmann::Solver::GMRES(LBSGroupset& groupset,
...
+  //================================================== Get groupset dof sizes
+  size_t groupset_numgrps = groupset.groups.size();
+  auto num_delayed_ang_unknowns = groupset.angle_agg.GetNumberOfAngularUnknowns();
+  size_t local_size = local_node_count * num_moments * groupset_numgrps +
+                      num_delayed_ang_unknowns.first;
+  size_t globl_size = glob_node_count * num_moments * groupset_numgrps +
+                      num_delayed_ang_unknowns.second;
+
+  //================================================== Create PETSc vectors
+  phi_new = chi_math::PETScUtils::CreateVector(static_cast<int64_t>(local_size),
```
wdhawkins, line 56: Why are we casting these values to int64?

ChiModules/LinearBoltzmannSolver/IterativeMethods/lbs_GMRES.cc
```diff
@@ -40,65 +44,53 @@ bool LinearBoltzmann::Solver::GMRES(LBSGroupset& groupset,
...
-  auto num_ang_unknowns = groupset.angle_agg.GetNumberOfAngularUnknowns();
-  int local_size = local_dof_count*num_moments*groupset_numgrps +
-                   num_ang_unknowns.first;
-  int globl_size = glob_dof_count*num_moments*groupset_numgrps +
-                   num_ang_unknowns.second;
-  MatCreateShell(PETSC_COMM_WORLD,local_size,
-                                  local_size,
-                                  globl_size,
-                                  globl_size,
+  MatCreateShell(PETSC_COMM_WORLD,static_cast<int64_t>(local_size),
```
wdhawkins, line 72: As above, why is this cast necessary?

ChiModules/LinearBoltzmannSolver/IterativeMethods/lbs_GMRES.cc
```diff
@@ -185,8 +180,8 @@ bool LinearBoltzmann::Solver::GMRES(LBSGroupset& groupset,
       chi_log.ProcessEvent(source_event_tag,
                            ChiLog::EventOperation::AVERAGE_DURATION);
     size_t num_angles = groupset.quadrature->abscissae.size();
-    long int num_unknowns = (long int)glob_dof_count*
-                            (long int)num_angles*
+    long int num_unknowns = (long int)glob_node_count *
```
wdhawkins: Since I believe these are all size_t, why the casts?

ChiModules/LinearBoltzmannSolver/IterativeMethods/lbs_GMRES.cc
```diff
@@ -204,7 +199,8 @@ bool LinearBoltzmann::Solver::GMRES(LBSGroupset& groupset,
         << chunk_overhead_ratio;
       chi_log.Log(LOG_0)
         << "        Sweep Time/Unknown (ns):       "
-        << sweep_time*1.0e9*chi_mpi.process_count/num_unknowns;
+        << sweep_time*1.0e9*chi_mpi.process_count/
```
wdhawkins, line 202: Just a note that at some point, we really need to convert to chrono and duration_cast to avoid things like a random multiply by 1e9.

ChiModules/LinearBoltzmannSolver/Tools/ksp_data_context.h
```diff
@@ -1,16 +1,38 @@
...
-  int last_iteration = -1;
-};
\ No newline at end of file
+  LinearBoltzmann::Solver& solver;
+  SweepChunk&              sweep_chunk;
+  LBSGroupset&             groupset;
+  Vec&                     operating_vector;
+  chi_mesh::sweep_management::SweepScheduler& sweepScheduler;
+  SourceFlags    lhs_scope;
+  size_t last_iteration = -1;
```
wdhawkins: Not sure if this is intentional or not. While wraparound is well defined by the standard, I argue that it's bad practice.  If it should accept signed values, use a signed type.

ChiModules/LinearBoltzmannSolver/Tools/kspmonitor_npt.cc
```diff
@@ -81,28 +62,28 @@ KSPConvergenceTestNPT(KSP ksp, PetscInt n, PetscReal rnorm,
...
 
-  if (context->groupset->iterative_method == NPT_GMRES)
+  if (context->groupset.iterative_method == NPT_GMRES)
   {
     if (context->last_iteration == n)
     {
-      if (context->solver->options.write_restart_data)
+      if (context->solver.options.write_restart_data)
       {
         if ((chi_program_timer.GetTime()/60000.0) >
```
wdhawkins: Although not part of the pr, we should document the choice of 60000.0.

ragusa:
Good point. I wonder too.

—jean ragusa

multiphysics.engr.tamu.edu
nustem.engr.tamu.edu

On May 21, 2021, at 1:27 PM, wdhawkins ***@***.***> wrote:

﻿

@wdhawkins commented on this pull request.

________________________________

In ChiModules/LinearBoltzmannSolver/Tools/kspmonitor_npt.cc<https://urldefense.com/v3/__https://github.com/chi-tech/chi-tech/pull/131*discussion_r637121088__;Iw!!KwNVnqRv!VGkK47o9epY4KaQsgRN6DM8nXMDRB7PzErjRBqJCw18ftQjGdkZfIRlEr2ZchCNxWw$>:

>    {
     if (context->last_iteration == n)
     {
-      if (context->solver->options.write_restart_data)
+      if (context->solver.options.write_restart_data)
       {
         if ((chi_program_timer.GetTime()/60000.0) >


Although not part of the pr, we should document the choice of 60000.0.

ChiModules/LinearBoltzmannSolver/lbs_01d_initparrays.cc
```diff
@@ -55,14 +55,14 @@ void LinearBoltzmann::Solver::InitializeParrays()
...
-  glob_dof_count = discretization->GetNumGlobalDOFs(flux_moments_uk_man)/GxM;
+  auto& per_node = ChiMath::UNITARY_UNKNOWN_MANAGER;
+  local_node_count = discretization->GetNumLocalDOFs(per_node);
+  glob_node_count = discretization->GetNumGlobalDOFs(per_node);
 
   //================================================== Compute num of unknowns
   int num_grps = groups.size();
   int M = num_moments;
-  unsigned long long local_unknown_count = local_dof_count * num_grps * M;
+  unsigned long long local_unknown_count = local_node_count * num_grps * M;
```
wdhawkins, line 65: I think this has the potential to cause problems if num_grps * M exceeds int_max.

ChiModules/LinearBoltzmannSolver/IterativeMethods/lbs_GMRES.cc
Naktakala: Because that is what PETSc takes. On that note, we could create an issue to modify the `PETScUtils` so that those utility functions use `PetscInt` instead of `int64_t` (which would be more consistent). What do you think?

ChiModules/LinearBoltzmannSolver/lbs_03d_wgdsa.cc
```diff
@@ -110,7 +110,7 @@ void LinearBoltzmann::Solver::AssembleWGDSADeltaPhiVector(LBSGroupset& groupset,
   int gss = groupset.groups.size();
 
   delta_phi_local.clear();
-  delta_phi_local.resize(local_dof_count*gss,0.0);
+  delta_phi_local.resize(local_node_count * gss, 0.0);
```
wdhawkins, line 113: Just an interesting observation that resize followed by fill might be faster than clear and resize.

wdhawkins (changes requested):
My only major concern is with the wraparound in ksp_data_context.

ChiModules/LinearBoltzmannSolver/IterativeMethods/lbs_GMRES.cc
Naktakala: Fixed.

ChiModules/LinearBoltzmannSolver/Tools/ksp_data_context.h
Naktakala: Great catch Thx!

ChiModules/LinearBoltzmannSolver/lbs_01d_initparrays.cc
Naktakala: How to fix?

ChiModules/LinearBoltzmannSolver/lbs_03d_wgdsa.cc
Naktakala: What would you like to see here?

ChiModules/LinearBoltzmannSolver/Tools/kspmonitor_npt.cc
Naktakala: More descriptive constant defined. This was intended to be an input parameter. Will create issue. For now, this is limited to an interval of 60s (60k milliseconds).

ChiModules/LinearBoltzmannSolver/lbs_03d_wgdsa.cc
wdhawkins: No fix necessary.  Just an observation.

ChiModules/LinearBoltzmannSolver/IterativeMethods/lbs_GMRES.cc
wdhawkins: Yes, I think that's probably the correct thing to do.  It sounds like it's at least the consistent thing to do.

ChiModules/LinearBoltzmannSolver/lbs_01d_initparrays.cc
wdhawkins: If num_grps is size_t (which would be consistent since it comes from size()), that should fix it.

wdhawkins (approved):

Issue #133, opened 2021-05-21
make time-dependent portion of the code more descriptive/self-explanatory


ragusa:
I stubbled upon a "xs->J ". I've understood that J is the # of delayed n precurosrs groups (we used that in class ...). However, I think a more descriptive terminology should be used in the code (look at the number of groups for example).

This may just be the tip of the iceberg so please go through your code additions and make sure the variables have more explicit names throughout. Thanks

Naktakala:
Resolved during PR #134

PR #134, opened 2021-05-21
More specific nomenclature in TransportCrossSections.

zachhardy:
Changed the nomenclature in the TransportCrossSections class to be more informative.

G -> num_groups
L -> scattering_order
J -> num_precursors

This addresses issue #133.

### DIFF
--- a/ChiModules/LinearBoltzmannSolver/Tools/kspmonitor_npt.cc
+++ b/ChiModules/LinearBoltzmannSolver/Tools/kspmonitor_npt.cc
@@ -61,14 +61,14 @@ PetscErrorCode LinearBoltzmann::
 
 
   chi_log.Log(LOG_0) << iter_info.str() << std::endl;
-
+  const double SIXTY_SECOND_INTERVAL = 60000.0; //time in milliseconds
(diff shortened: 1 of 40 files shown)


ragusa (approved):

Naktakala (changes requested):
This is ready. It just needs a rebase on the latest development because there were changes to SetKSource

Naktakala (approved):

ChiModules/LinearBoltzmannSolver/Tools/kspmonitor_npt.cc
```diff
@@ -80,29 +61,30 @@ KSPConvergenceTestNPT(KSP ksp, PetscInt n, PetscReal rnorm,
 
 
   chi_log.Log(LOG_0) << iter_info.str() << std::endl;
-
-  if (context->groupset->iterative_method == NPT_GMRES)
+  const double SIXTY_SECOND_INTERVAL = 60000.0; //time in milliseconds
```
ragusa, line 64: how do we know the time was given in msec? 
also, isn't a sixty-second interval also a minute?  If so, why not call this factor: convert_to_minutes or something like that?

PR #139, opened 2021-05-31
Bug introduced during PR

Naktakala:
This was totally my fault. Normally I pull down a PR and compile it on 3 different compilers.


ragusa (commented):
I do not fully understand. Can't we have a map with both int and unsigned int?  It is clear that ell should be unsigned. Why did you have to make it an int ?

Naktakala:
> I do not fully understand. Can't we have a map with both int and unsigned int? It is clear that ell should be unsigned. Why did you have to make it an int ?

What happens is on the `for m` line the minus gets applied to the order of operations before the assignment happens. So `-ell` gets interpreted first and if `ell=1` (or greater) with type `uint` then `-uint` immediately loops back to a really large number and remains of type `uint`. This then doesn't pass the `<=ell` and the loop stops. It then only works for `ell=0` since `-(uint=0)` remains `uint=0` and essentially dropped all `L=1` simulations to `L=0` and also explains why the 1D sims were fine.

I like the simplicity of just having the `ints` here. Unless somebody can come up with an alternative not as ugly as this:

```c++
for (unsigned int ell=0; ell<=scattering_order; ++ell)
  for (int m=-static_cast<int>(ell); m<=static_cast<int>(ell); ++m)
    etc.
```

Sometimes I feel like things like `const` and `unsigned` unnecessarily adds fluff. Especially `unsigned` stuff can be dangerous for the case we have here.

ragusa:
I understand. Int everywhere is fine.

Naktakala:
Also. I like to see people use `int` when they are talking integers and counters. It just flows naturally. You don't really talk about unsigned integers in real life do you?. Cell-indices, sizes can definitely be used as `unsigned` and can add a little extra capacity.

I've always thought that loops like `for i, for j` just look silly with `unsigned` (like in deal.ii tutorials) because the scope is right there. When the loops have been verified to work then its unlikely that people can use those variables outside the loops. But that is just my opinion.

PETSc uses `int64_t` as they're `PetscInt` and avoids `unsigned` across the board.

wdhawkins:
My two cents... albeit a bit late.  I agree with Jan about using int.  In fact... with the exception of global indices (vertices, faces, etc.), we should use int everywhere else.  It's natural, cleaner, and the support for negative values can be useful. It also avoids problems with loops that decrement.

PR #153, opened 2021-06-22
Mesh quality bug

Naktakala:
A while ago we encountered a problem with importing tetrahedra from VTU. I made a change to reorient bad cells but this was a bad move since the bug was in the external code generating the VTU-file. Now there is only a check for negative volume triangles or tetrahedra.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -102,7 +102,7 @@ else ()
     )
 endif()
 
-set(CHI_LIBS lua m dl ${MPI_CXX_LIBRARIES} petsc ${VTK_LIBRARIES})
+set(CHI_LIBS stdc++ lua m dl ${MPI_CXX_LIBRARIES} petsc ${VTK_LIBRARIES})
 
 #================================================ Compiler flags
 if(CMAKE_CXX_COMPILER_ID STREQUAL "Clang")
--- a/ChiTech/ChiMesh/UnpartitionedMesh/unpartmesh_00a_general.cc
+++ b/ChiTech/ChiMesh/UnpartitionedMesh/unpartmesh_00a_general.cc
@@ -11,6 +11,8 @@ extern ChiTimer chi_program_timer;
 /**Compute centroids for all cells.*/
 void chi_mesh::UnpartitionedMesh::ComputeCentroidsAndCheckQuality()
 {
+  const chi_mesh::Vector3 khat(0.0,0.0,1.0);
+
   chi_log.Log() << "Computing cell-centroids.";
   for (auto cell : raw_cells)
   {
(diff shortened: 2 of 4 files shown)


CMakeLists.txt
```diff
@@ -102,7 +102,7 @@ else ()
     )
 endif()
 
-set(CHI_LIBS lua m dl ${MPI_CXX_LIBRARIES} petsc ${VTK_LIBRARIES})
+set(CHI_LIBS stdc++ lua m dl ${MPI_CXX_LIBRARIES} petsc ${VTK_LIBRARIES})
```
Naktakala, line 105: I threw this in here because it seems some compiler configurations on TACC wants this to be there. It should be harmless.

ChiTech/ChiMesh/UnpartitionedMesh/unpartmesh_00a_general.cc
```diff
@@ -11,55 +11,81 @@ extern ChiTimer chi_program_timer;
...
+    }//if polyhedron
+  }//for cell in raw_cells
 
-  if (check0_corrections > 0)
+  if (num_negative_volume_elements > 0)
     chi_log.Log(LOG_ALLWARNING)
-      << "Cell-center-to-face orientation detected " << check0_corrections
-      << " faces that violated quality requirements. An attempt to fix it"
-      << " was made.";
+      << "Cell-center-to-face orientation detected " << num_negative_volume_elements
```
ragusa: A little unsure of the meaning of cell-center-to-face orientation. What is meant by this?

ragusa (approved):
I left one comment

ChiTech/ChiMesh/UnpartitionedMesh/unpartmesh_00a_general.cc
Naktakala: Mmmm. Help me find better words for this. When a cell-centroid looks at a face then the face should be oriented away from the centroid in order to be able to form a positive volume triangle/tet. In other words, forming a vector from the cell-centroid to the face-centroid (or any of its vertices for that matter), then the dot product with the face-normal `n` should be positive.

CMakeLists.txt
dominiccaron: Can you include that modification in ```Downstream.cmake``` as well?

dominiccaron (changes requested):

CMakeLists.txt
Naktakala: Done

ChiTech/ChiMesh/UnpartitionedMesh/unpartmesh_00a_general.cc
ragusa: I would be explicit:
- I guess it necessary applies to tri/tet for now so let's say in the error message that there is a tri/tet orientation issue
- maybe add " issue" --> orientation issue detected


alsom is this specific to PWL discretization? Would there be an issue if LD was used?

dominiccaron (approved):

ChiTech/ChiMesh/UnpartitionedMesh/unpartmesh_00a_general.cc
Naktakala: Reworded. Message now reads like this:

```
Cell quality checks detected 3 negative volume
 sub-elements (sub-triangle or sub-tetrahedron). 
This can create issues for some spatial discretizations.
```

PR #152, opened 2021-06-21
VTK utilities modifications for VTK 9 support take 2

Naktakala:
## Highlights
- The main `CMakeLists.txt` has been updated to handle both VTK <8.9 and >8.9
- The export field-function routines have been updated to do what is compatible with both VTK ranges (this mostly involved the way polyhedron-faces were exported).
- Some minor changes of `int` to larger types
- The routines `ExportToVTKG` used to export all the components of an unknown whilst `ExportToVTK` exported only a single component. They have now been merged into just `ExportToVTK`

## Future plans that this PR reveals
- Field-functions need to be revamped. 
  - They need to be coupled to an `Unknown` (i.e. SCALAR, VECTOR_N, or TENSOR) and not to a specific component
  - They need to take ownership of their data-vector. This is creating confusion since they are now mapping into a multi-unknown data-vector owned by the solvers. Solvers need to "transact" with a field-function via something like an `Update` function which will supply a compacted data-vector for a specific unknown only.
  - Exporting multiple field-functions into one .pvtu requires the same grid AND the same spatial discretization. This can be changed to just requiring the same grid, and then used to great effect to support additional functionality.
  - Field-functions are now passed around via handles (which are integers). This needs to be expanded to get the handles from text names and allow for more readable input and coding.
  - One needs to be able to map one field-function to another without the grid or spatial discretization being the same (yum yum coarse grid acceleration, multi-physics, mmmmm).

PR #154, opened 2021-06-23
Setsource refurbish 1

zachhardy:
## Highlights
- Some rearrangement of the `KEigenvalueSolver`.
- General verbosity of messages changed.
- Retooling of `SetSource` in `LBS`.

### `KEigenvalueSolver` changes
- Numerous rearrangements to the flow of a K-Eigen solve. Most of it is geared towards supporting multiple groupsets which the solver currently does not support.
- The `KEigenvalueSolver` now has `PowerIteration` as the only outer method and can choose to use either `ClassicRichardson` or `GMRES` for inner iterations (converging scattering). The source moments are initialized with an initial fission-only source and then the standard LBS algorithms are used. PS: This is quite a novel change.

### LBS changes
- The `SetSource` function now operates on a reference `q_moments` vector instead of the class member vector `q_moments_local` making the manipulation of sources a lot better. i.e. Since the LBS steady state algorithms use `SetSource`, the new changes allows `GMRES` and `ClassicRichardson` to be used in the inner iterations of other algorithms.
- Some source flags have been added making Across-group set and Within-groupset consistent for fission. We don't know if this will be necessary yet but right now it doesn't hurt.


ChiModules/KEigenvalueSolver/IterativeOperations/set_k_source.cc
```diff
@@ -43,95 +46,130 @@ void KEigenvalue::Solver::
...
+            //============================== Across-groupset
+            if (apply_wgs_fission_src) {
+              //============================== Loop over groups
+              for (size_t gprime = first_grp; gprime <= last_grp; ++gprime) {
+                double nu_sig_f = (options.use_precursors) ?
+                                  xs->nu_prompt_sigma_f[gprime] : xs->nu_sigma_f[gprime];
+
+                if ((gprime >= gs_i) and (gprime <= gs_f)) {
+                  fission_g += xs->chi[g] * nu_sig_f *
+                               phi_prevp[gprime] / k_eff;
```
ragusa: what is the last p for in "phi_prevp" ??

ChiModules/KEigenvalueSolver/initialize_precursors.cc
```diff
@@ -15,46 +15,45 @@ using namespace LinearBoltzmann;
...
+      // ----- Cell information
+      const auto xs_id = matid_to_xs_map[cell.material_id];
+      auto &xs = material_xs[xs_id];
+      auto cell_fe_view = pwl->GetCellMappingFE(cell.local_id);
+      auto &transport_view = cell_transport_views[cell.local_id];
+
+      // ----- Loop over cell dofs
+      for (int i = 0; i < cell_fe_view->num_nodes; ++i) {
+        size_t ir = transport_view.MapDOF(i, 0, 0);
+        int64_t jr = pwl->MapDOFLocal(cell, i, Nj_unk_man, 0, 0);
```
ragusa: What is Nj_unk_man?
Can you add a comment where that var is defined? If I do not know immediately what it stands for, maybe others in the future will wonder too. Thanks

ChiModules/KEigenvalueSolver/initialize_precursors.cc
```diff
@@ -15,46 +15,45 @@ using namespace LinearBoltzmann;
...
+          for (size_t j = 0; j < xs->num_precursors; ++j) {
+            size_t j_map = precursor_map[xs_id][j];
+
+            // ----- Initialize precursors
+            double coeff = xs->precursor_yield[j] / xs->precursor_lambda[j];
+
+
+            // ----- Loop over groups
+            for (int g = 0; g < groups.size(); ++g)
+              Nj_newp[j_map] += coeff * xs->nu_delayed_sigma_f[g] *
```
ragusa: here too there's a p at the end of the varname. What is it for?

ChiModules/KEigenvalueSolver/initialize_precursors.cc
```diff
@@ -15,46 +15,45 @@ using namespace LinearBoltzmann;
...
+                                phi_newp[g] / k_eff;
+          }//for j
+        }//if fissile and J > 0
```
ragusa: J ? change to actual name

ChiModules/KEigenvalueSolver/IterativeOperations/set_k_source.cc
```diff
@@ -43,95 +46,130 @@ void KEigenvalue::Solver::
...
+
+          q_mom[g] += inscatter_g;
+
+          //======================================== Apply fission
+          if (xs->is_fissile and (ell == 0)) {
+            double fission_g = 0.0;
+            //============================== Across-groupset
+            if (apply_ags_fission_src) {
+              //============================== Loop over groups
+              for (size_t gprime = first_grp; gprime <= last_grp; ++gprime) {
```
ragusa: I am used to seeing C++ loops for var < var_max but here it is var <= var_max. Why <= here (and later too)?

ChiTech/ChiMesh/MeshContinuum/chi_meshcontinuum_exportvtk.cc
```diff
@@ -131,12 +121,11 @@ void chi_mesh::MeshContinuum::ExportCellsToVTK(const char* baseName)
         d_node[1] = grid->vertices[vgi].y;
         d_node[2] = grid->vertices[vgi].z;
 
-        points->InsertPoint(nc,d_node.data());
-        cell_info[v] = nc; nc++;
+        points->InsertPoint(node_count, d_node.data());
+        cell_info[v] = node_count++;
```
ragusa: ++node_count or node_count++? and why??

ChiTech/ChiPhysics/FieldFunction/export_multiple_vtk.cc
```diff
@@ -75,7 +73,7 @@ void chi_physics::FieldFunction::
 
   //============================================= Populate VTK points
   auto points = vtkSmartPointer<vtkPoints>::New();
-  int vc=0;
+  int64_t vc=0;
```
ragusa: @Naktakala I think there is another PR where very short varnames have been made more explicit. I think I remember this to be the case for "vc". If I am correct, then we should try to remain consistent throughout. Thanks

ragusa (commented):
just a few comments

ChiModules/KEigenvalueSolver/IterativeOperations/set_k_source.cc
zachhardy: This is the syntax used in the SetSource routine within the LinearBoltzmann::Solver class, I simply mirrored that here. What the last "p" is for is best answered by @Naktakala .

ChiModules/KEigenvalueSolver/initialize_precursors.cc
zachhardy: This is the unknown manager for the delayed neutron precursors, named Nj because that is what the delayed precursor vector is named currently. I added a comment in the header file to explain that 'Nj' is the precursors. I just changed 'unk_man' to 'uk_man' to be consistent with unknown manager naming elsewhere. We could change 'Nj_...' to precursor_...' for more readability. Thoughts?

ragusa: I am always for more clarity/readability. I do not think we have have varname with 200 characters but there is likely a good trade-off somewhere.
@Naktakala So uk is like a dof_handler in deal.ii?

ChiModules/KEigenvalueSolver/initialize_precursors.cc
zachhardy: Done.

ChiModules/KEigenvalueSolver/IterativeOperations/set_k_source.cc
zachhardy: This is how it is done in the SetSource routine in the LinearBoltzmann::Solver class. This loops from the id of the first group in the list of group objects to the id of the last group in the list of group objects. Doing this does suggest that the first group has to be smallest id and the last group the largest. I am unsure why this was the decision made within the LinearBoltzmann::Solver class as opposed to looping from 0 -> num_groups. @Naktakala could probably answer this.

ChiTech/ChiMesh/MeshContinuum/chi_meshcontinuum_exportvtk.cc
zachhardy: I think this is here because this PR was based on the #152, which was just merged.

ChiModules/KEigenvalueSolver/IterativeMethods/power_iteration.cc
```diff
@@ -25,161 +28,128 @@ using namespace LinearBoltzmann;
...
-      }
-    }//for source iterations
-
-    if (!si_converged) {
-      chi_log.Log(LOG_ALLVERBOSE_1)
-          << "\n!!!WARNING!!! "
-             "Source iterations did not converge.\n";
+      ClassicRichardson(groupset, 0, sweep_scheduler,
+                        APPLY_WGS_SCATTER_SOURCE | APPLY_AGS_SCATTER_SOURCE,
+                        false);
```
jthano: I think it might be good to change `false` to
```c++
options.verbose_inner_iterations
```
Same thing for the GMRES function

ChiModules/KEigenvalueSolver/IterativeMethods/power_iteration.cc
```diff
@@ -25,161 +28,128 @@ using namespace LinearBoltzmann;
...
-        break;
-      }
-    }//for source iterations
-
-    if (!si_converged) {
-      chi_log.Log(LOG_ALLVERBOSE_1)
-          << "\n!!!WARNING!!! "
-             "Source iterations did not converge.\n";
+      ClassicRichardson(groupset, 0, sweep_scheduler,
+                        APPLY_WGS_SCATTER_SOURCE | APPLY_AGS_SCATTER_SOURCE,
```
jthano, line 75: I think that the `APPLY_MATERIAL_SOURCE` flag needs to be added here. Even though there isn't any fixed source, per #156 I think we won't have reflecting boundaries without a `APPLY_MATERIAL_SOURCE` flag

ChiModules/KEigenvalueSolver/IterativeMethods/power_iteration.cc
Naktakala: I agree with this.

ChiModules/KEigenvalueSolver/IterativeMethods/power_iteration.cc
ragusa: @Naktakala : at some point, would it be possible (if it does not yet exist) to have a short document (even an .md) describing the various APPLY_xxxx, their obvious functions but also their maybe-less-obvious functions as well?
[what @jthano is describing in his comments falls into the less-obvious, to me at least]

Naktakala: I was actually alarmed by this @jthano because I initially thought it is not correct. The `APPLY_MATERIAL_SOURCE` flag is used with a sweep-chunk setting to determine whether incident boundary conditions are to be applied during a sweep. However, if the boundary is reflecting then this flag needs to be ignored. Surprise: It's not being ignored as can be seen here

```c++
double* chi_mesh::sweep_management::AngleSet::
PsiBndry(uint64_t bndry_map,
         int angle_num,
         uint64_t cell_local_id,
         int face_num,
         int fi,
         int g,
         int gs_ss_begin,
         bool surface_source_active)
{
 if (not surface_source_active)
    return &ref_boundaries[bndry_map]->zero_boundary_flux[g];

  if (ref_boundaries[bndry_map]->IsReflecting())
    return ref_boundaries[bndry_map]->HeterogenousPsiIncoming(
      angle_num, cell_local_id, face_num, fi, gs_ss_begin);

  return &ref_boundaries[bndry_map]->boundary_flux[g];
}
```

Now ... I am aggressively confident that somewhere in the past I changed this function and induced this bug. I shall fix it and the underlying lack of a regression which should've pointed this out as it only affects GMRES.

jthano: I have my C5G7 benchmark up and running, but I'm not getting the correct eigenvalue. This issue with reflecting boundaries is one thing I ran into while trying to get my benchmark working. I actually thought that #156 was specifically to fix issues related to reflecting boundaries.

ChiModules/KEigenvalueSolver/IterativeMethods/power_iteration.cc
```diff
@@ -25,161 +28,128 @@ using namespace LinearBoltzmann;
...
-    chi_log.Log(LOG_0VERBOSE_2)
-      << "\n********** Starting source iterations";
 
+    //============================== Clear source moments
+    q_moments_local.assign(q_moments_local.size(), 0.0);
+
+    //============================== Set the fission source
+    SetKSource(groupset, q_moments_local,
+               APPLY_AGS_FISSION_SOURCE |
+               APPLY_WGS_FISSION_SOURCE);
```
Naktakala, line 68: For those who don't understand this change. `SetKSource` here preloads the fission source with the k-division into `q_moments_local` which now removes the need to have a custom `SetSource` for the inner iterations because `SetSource` will simply add to `q_moments_local`.

The changes in `ClassicRichardson` and `GMRES` also restores `q_moments_local` to its state before they were called. In other words:
- `PowerIteration` iteration
  - `SetKSource`, `Q=F`
  - `Richardson` or `GMRES`
    - `Q_temp = Q`
    - `SetSource`, `Q=F+S+others`
    - Stuff
    - `Q = Q_temp`

Naktakala (approved):

ChiModules/KEigenvalueSolver/IterativeMethods/power_iteration.cc
jthano: Since the k fission source is simply preloaded into the q_moments_local vector, do we still need a third scalar flux vector? It seems like we can just use either `phi_old` or `phi_new` to first compute a fission integral, add a fission source to q_moments, converge a new iterate of 'phi_new', and finally compute a new fission integral to update `k`.

Also, should all the code relate to wgs scattering and ags scattering simply be removed from `SetKSource` since that function isn't meant to update scattering terms?

zachhardy: @Naktakala I would revise this to be:

- `PowerIteration` iteration
    -  `SetKSource` -> `Q = F`
    - `Q_temp = Q`
    - `Richardson` or `GMRES` iterations
        -  `Q = Q_temp`
        - `SetSource` -> `Q = F + S + others`
        - Stuff

We store `Q_temp` before the inner iterations and then initialize `Q = Q_temp` in the inner iterations.

zachhardy: @jthano This is the direction I want to take it, yes. I will likely make that change in the following PR which adds in groupsets to the k-eigenvalue solver. The primary purpose of this PR was to make the various `SetSource` routines write to a specified vector instead of clearing vectors within the routines.

ragusa: @Naktakala @jthano @zachhardy @dominiccaron @wdarylhawkins 
I like this discussion because I am learning something. Here's my thinking, which could be a concern. Those flags feel a little bit like a hack to me: "this is what happens to the source, provided that you have passed certain flags". I am afraid this can also be a source of errors or at least confusion. Is this the appropriate way of doing this?

wdhawkins: I agree with @ragusa. The flags infrastructure seems a bit error-prone to me. I have to admit though that I haven't had a chance to dig too deeply into the code, so I can't offer any constructive suggestions on what we might change or improve.  I'll make some time in the next day or so to take a look at things.

ChiModules/KEigenvalueSolver/IterativeMethods/power_iteration.cc
zachhardy: Done.

ChiModules/KEigenvalueSolver/IterativeMethods/power_iteration.cc
Naktakala: Mmmmmm. Disappointing to see that it causes confusion only now because of the reflective BCs. Apart from the bug with BCs, the intent of using these flags was originally to help others reuse pieces of code, i.e., `SetSource`. And to that end, I believe it was kinda successful for these situations:
- The boolean options for GMRES simplified a lot, reminder GMRES calls `SetSource` in 2 places:
  - First to compute a source driven by material/boundary sources + Across-GroupSet (AGS) scattering
  - Then to compute a source driven by Within-GroupSet (WGS) scattering from a Krylov-vector
- Classic Richardson and GMRES became available to use by the KEigenvalue solver.
- Overloading `SetSource` became easier. You could add your own flags and both ClassicRichardson and GMRES didn't care.

I would like to hear from @jthano and @zachhardy on this topic since I believe they were the intended clients. Was the bit-wise flags a design mistake?

ragusa: I think this is something I'd like to discuss on Friday.

jthano: I think the bit-wise flags are a good solution, they seem to be a popular and common tool for situations like this. I don't think the flags themselves are error prone, in fact I think they make the code more readable and less error prone.

It seems like there was just an odd code bug very specifically limited to reflecting boundaries that also happened to tangentially involve the flags.

zachhardy: Personally, I like the bit-wise flags as well. Having many routines that do essentially the same thing as another with a tiny change (like dividing by k in the fission source) is suboptimal and the only other alternative I see to this is having several functions for the various components of the source. At the end of the day, those function names would incorporate the same information as the bit-wise flags. I say keep them.

ragusa: who said we should have many routines that are similar :-) ? As of now, I think I mostly need to find in the document what the list of flags are and make sure everybody is well educated of their existence and then use

zachhardy: @ragusa Nobody said we had to but at the moment between LBS, the k-Eigenvalue Solver, and my transient solver (not in this repo) that is exactly what has happened. The bit-wise flags    have certainly taken us a step in the right direction away from duplicated code with minor changes.

Naktakala:
@dominiccaron @ragusa @wdhawkins I don't see any hard objections or required code changes. Can we move forward?

ragusa:
@dominiccaron @wdarylhawkins @Naktakala : 
- I think I better understand the bitwise flag now. At this stage, as I wrote in an earlier comment, I would like to see where I can quickly see the whole list of bitwise flags, and a short description. Can you point me to that, if it exists?
- I would not mind hearing back from Dominic and Daryl on these, just to have their opinion. We heard from Daryl I think but it was after a cursory look. I think one more look would be good.

ChiModules/KEigenvalueSolver/initialize_precursors.cc
```diff
@@ -15,45 +15,48 @@ using namespace LinearBoltzmann;
 /**Computes the point wise delayed neutron precursor concentrations.*/
 void KEigenvalue::Solver::InitializePrecursors()
 {
-  typedef SpatialDiscretization_PWLD  PWLD;
-  auto pwl = std::static_pointer_cast<PWLD>(discretization);
+  if (options.use_precursors)
+  {
+    typedef SpatialDiscretization_PWLD  PWLD;
+    auto pwl = std::static_pointer_cast<PWLD>(discretization);
```
dominiccaron: As above.

ChiModules/KEigenvalueSolver/initialize_k_solver.cc
```diff
@@ -10,38 +11,36 @@ using namespace LinearBoltzmann;
...
+      if (xs->num_precursors > max_num_precursors_per_material)
+        max_num_precursors_per_material = xs->num_precursors;
     }
 
-    // ----- Initialize precursor unknown manager and vector
+
     if (num_precursors > 0)
     {
       typedef SpatialDiscretization_PWLD  PWLD;
       auto pwl = std::static_pointer_cast<PWLD>(discretization);
```
dominiccaron: Notwithstanding that LBS currently is fully implemented for PWLD spatial discretisations only, I do not believe that this method is (or need be) called at a location of the code where it is known that the spatial discretisation is necessarily of type PWLD. If correct, then it may be beneficial to add a verification here to prevent against a bad downcast so that eventual unexpected behaviour is identified easily in the future if/when other spatial discretisations enter into the code.

ChiModules/KEigenvalueSolver/IterativeMethods/power_iteration.cc
```diff
@@ -25,161 +28,128 @@ using namespace LinearBoltzmann;
...
-      << "  k_eff change " << std::setw(10) << k_eff_change
-      << "  reactivity " << std::setw(10) << reactivity * 1e5;
-    if (k_converged) {
-      k_iter_info << " CONVERGED\n";
+    double k_eff_change = fabs(k_eff - k_eff_prev) / k_eff;
+    k_eff_prev = k_eff;
+    F_prev = F_new;
+    nit += 1;
+
+    if (k_eff_change < std::max(options.tolerance, 1.0e-12))
```
dominiccaron, line 97: For general thought: personally, I am not a proponent of limiting user-requested calculation parameters in this manner, especially silently inside the solver. There may be valid reasons for requesting very tight convergence tolerances; this approach effectively limits what is achievable. However, if such limiting must occur, it is good practice to notify the user where/when it occurs. Is this approach adopted elsewhere in the code?

ChiModules/KEigenvalueSolver/IterativeMethods/power_iteration.cc
dominiccaron: My thoughts: there exist multiple ways to implement this particular functionality. In all approaches that come to mind, some degree of control over which components of the source are superimposed is required. The bit-wise flags provide the necessary information in a concise form. In my mind, the processing of those options in a manner that does not lead to error or to duplicate methods/code is not necessarily coupled to the use of the flags construct itself.

ChiModules/LinearBoltzmannSolver/lbs_linear_boltzmann_solver.h
```diff
@@ -30,11 +30,12 @@ enum class BoundaryType
...
-    NO_FLAGS_SET           = 0,
-    APPLY_MATERIAL_SOURCE  = (1 << 0),
+    NO_FLAGS_SET               = 0,
+    APPLY_MATERIAL_SOURCE      = (1 << 0),
     APPLY_WGS_SCATTER_SOURCE   = (1 << 1),
     APPLY_AGS_SCATTER_SOURCE   = (1 << 2),
-    APPLY_FISSION_SOURCE       = (1 << 3)
+    APPLY_WGS_FISSION_SOURCE   = (1 << 3),
+    APPLY_AGS_FISSION_SOURCE   = (1 << 4)
   };
```
dominiccaron, line 39: As mentioned by others, this needs to be documented.

dominiccaron (changes requested):
A few minor comments and an observation for general consideration.

ChiModules/KEigenvalueSolver/IterativeMethods/power_iteration.cc
zachhardy: Within the ClassicRichardson method this is used: 

`if (pw_change<std::max(groupset.residual_tolerance*(1.0-rho),1.0e-10))`

which was the primary reason this has been here since the adoption of the `KEigenvalue::Solver` module. Perhaps make an issue for this to be addressed later.

ChiModules/LinearBoltzmannSolver/lbs_linear_boltzmann_solver.h
zachhardy: This is true, but these were accepted without documentation prior to this PR in one authored not by myself. The adoption of the bit-wise flags is not really within the scope of this PR and I think would be better addressed by raising an issue.

ChiModules/KEigenvalueSolver/initialize_k_solver.cc
zachhardy: This has now been removed. The `SpatialDiscretization` class has the method available that we use in this chunk of code, so now it simply calls `discretization->GetNumLocalDOFs(precursor_uk_man)` to get the number of dofs to size the vector.

ChiModules/KEigenvalueSolver/IterativeMethods/power_iteration.cc
wdhawkins: I agree completely agree with Dominic.  I'm not at all opposed to the bit-wise flag infrastructure. It may very well turn out that we all agree that it's the best implementation for this functionality. I do think we should give it careful consideration since it underpins a number of features.

ChiModules/KEigenvalueSolver/IterativeMethods/power_iteration.cc
```diff
@@ -25,161 +28,128 @@ using namespace LinearBoltzmann;
...
+      std::stringstream k_iter_info;
+      k_iter_info
+          << chi_program_timer.GetTimeString() << " "
+          << "  Iteration " << std::setw(5) << nit
+          << "  k_eff " << std::setw(10) << k_eff
+          << "  k_eff change " << std::setw(10) << k_eff_change
+          << "  reactivity " << std::setw(10) << reactivity * 1e5;
+      if (converged) k_iter_info << " CONVERGED\n";
+
+      chi_log.Log(LOG_0) << k_iter_info.str();
```
wdhawkins, line 112: This comment probably just demonstrates my unfamiliarity with chi_log, but do we need the stringstream here?  Does:

chi_log.log(LOG_0) << chi_program_time.GetTimeString() << " " << " Iteration....

not work?

ChiModules/KEigenvalueSolver/IterativeOperations/compute_production.cc
```diff
@@ -30,17 +31,17 @@ double KEigenvalue::Solver::ComputeProduction()
     auto& transport_view = cell_transport_views[cell.local_id];
 
     // ----- Loop over cell dofs
-    for (int i=0; i<cell_fe_view.NumNodes(); i++)
+    for (int i = 0; i < cell_fe_view.NumNodes(); i++)
```
wdhawkins: Just being pedantic here, but could we convert i++ to ++i (same in the loops that follow)?

ChiModules/KEigenvalueSolver/IterativeOperations/set_k_source.cc
```diff
@@ -41,102 +44,160 @@ void KEigenvalue::Solver::
...
   {
     auto& full_cell_view = cell_transport_views[cell.local_id];
 
-    // ----- Obtain cross-section and src
+    //==================== Obtain cross-section and src
     int cell_matid = cell.material_id;
     int xs_id = matid_to_xs_map[cell_matid];
 
-    if ( (xs_id<0) || (xs_id>=material_xs.size()) )
+    if ((xs_id < 0) || (xs_id >= material_xs.size()))
```
wdhawkins, line 58: I'm surprised this didn't generate a compiler warning, but we should either promote xs_id to size_t or downcast material_xs.size().

ChiModules/KEigenvalueSolver/IterativeOperations/set_k_source.cc
```diff
@@ -41,102 +44,160 @@ void KEigenvalue::Solver::
...
     {
       chi_log.Log(LOG_ALLERROR)
-      << "Cross-section lookup error\n";
+          << "Cross-section lookup error\n";
       exit(EXIT_FAILURE);
```
wdhawkins, line 62: Not really part of the PR, but IMHO, calling exit() in an MPI code is bad form.  We really should be using MPI_Abort() (at least if we're not going to try to clean things up ourselves and we're not in a signal handler). This should probably be opened as a separate issue. Calling exit() will terminate this process, but it may also result in hung processes on this (and other) nodes.  MPI_Abort() will at least attempt to prevent that.

ChiModules/KEigenvalueSolver/IterativeOperations/set_k_source.cc
```diff
@@ -41,102 +44,160 @@ void KEigenvalue::Solver::
...
-            {
-              for (size_t gprime=first_grp; gprime<=last_grp; ++gprime)
-                if (options.use_precursors)
-                  fission_g += xs->chi[g] * xs->nu_prompt_sigma_f[gprime] *
-                               phi_prevp[gprime] / k_eff;
-                else
-                  fission_g += xs->chi[g] * xs->nu_sigma_f[gprime] *
-                               phi_prevp[gprime] / k_eff;
+              //============================== Loop over transfers
+              for (int t = 0; t < num_transfers; ++t)
```
wdhawkins: Another signed/unsigned comparison. t should be size_t.

ChiModules/KEigenvalueSolver/IterativeOperations/set_k_source.cc
```diff
@@ -41,102 +44,160 @@ void KEigenvalue::Solver::
...
-            if ((apply_mat_src) and (xs->num_precursors > 0))
+            //============================== Within-groupset
+            if (apply_wgs_scatter_src)
             {
-              for (int j=0; j<num_precursors; ++j)
+              size_t num_transfers =
+                  xs->transfer_matrices[ell].rowI_indices[g].size();
+
+              //============================== Loop over transfers
+              for (int t = 0; t < num_transfers; ++t)
```
wdhawkins: t need to be size_t.

ChiModules/KEigenvalueSolver/initialize_precursors.cc
```diff
@@ -15,45 +15,48 @@ using namespace LinearBoltzmann;
...
+      auto& xs = material_xs[xs_id];
+      auto cell_fe_view = pwl->GetCellMappingFE(cell.local_id);
+      auto& transport_view = cell_transport_views[cell.local_id];
+
+      //============================== Loop over nodes
+      for (int i = 0; i < cell_fe_view->num_nodes; ++i)
+      {
+        size_t ir = transport_view.MapDOF(i, 0, 0);
+        size_t jr = pwl->MapDOFLocal(cell, i, precursor_uk_man, 0, 0);
+        double* Nj_newp = &precursor_new_local[jr];
```
wdhawkins, line 36: Just my opinion, but I think this is unnecessary and more convoluted than just adding an offset below. I also doubt that it's faster.

ChiModules/LinearBoltzmannSolver/IterativeOperations/lbs_setsource.cc
```diff
@@ -20,136 +20,152 @@ extern ChiLog& chi_log;
...
-    //=========================================== Obtain cross-section and src
+    //==================== Obtain xs and src
     int cell_matid = cell.material_id;
-
     int xs_id = matid_to_xs_map[cell_matid];
-    int src_id= matid_to_src_map[cell_matid];
+    int src_id = matid_to_src_map[cell_matid];
 
-    if ((xs_id<0) || (xs_id>=material_xs.size()))
+    if ((xs_id < 0) || (xs_id >= material_xs.size()))
```
wdhawkins, line 57: signed/unsigned comparison

ChiModules/LinearBoltzmannSolver/IterativeOperations/lbs_setsource.cc
```diff
@@ -20,136 +20,152 @@ extern ChiLog& chi_log;
...
+          //============================== Apply across-groupset scattering
+          if (moment_avail and apply_ags_scatter_src)
           {
-            size_t num_transfers = xs->transfer_matrices[ell].rowI_indices[g].size();
-            for (int t=0; t<num_transfers; t++)
+            size_t num_transfers =
+                xs->transfer_matrices[ell].rowI_indices[g].size();
+
+            //============================== Loop over transfers
+            for (int t = 0; t < num_transfers; ++t)
```
wdhawkins: signed/unsigned comparison.  I'll quit noting them going forward.

wdhawkins (changes requested):

ChiModules/KEigenvalueSolver/IterativeOperations/compute_production.cc
zachhardy: Done.

ChiModules/KEigenvalueSolver/IterativeMethods/power_iteration.cc
zachhardy: This is done this way to follow the syntax of the ClassicRichardson routine. I believe the Log routine moves to a new line in each successive call. Because the "CONVERGED\n" message is intended to be on the same line as the iteration summary, calling Log(LOG_0) for the summary and then for the convergence message would change this behavior. This is just my take on why I think this is used. As I said, it was not originally my design, I just used the same syntax so the code reads similarly.

ChiModules/KEigenvalueSolver/IterativeOperations/set_k_source.cc
zachhardy: Perhaps we should raise an issue for this. This is the exact syntax used everywhere else in the code cross-sections are checked and is not a design introduced by this PR.

ChiModules/KEigenvalueSolver/initialize_precursors.cc
zachhardy: Perhaps. This is a design that has been used since I began using ChiTech that is used in more places than this PR affects. I would suggest raising a separate issue for this.

ChiModules/LinearBoltzmannSolver/IterativeOperations/lbs_setsource.cc
zachhardy: See above.

ChiModules/LinearBoltzmannSolver/IterativeOperations/lbs_setsource.cc
zachhardy: Fixed.

ChiModules/KEigenvalueSolver/IterativeOperations/set_k_source.cc
zachhardy: Fixed.

ChiModules/KEigenvalueSolver/IterativeOperations/set_k_source.cc
zachhardy: Fixed.

ChiModules/KEigenvalueSolver/IterativeOperations/set_k_source.cc
zachhardy: Definitely agree on the separate issue. The exit call is used across the code and was not a design introduced in this PR.

ragusa:
@wdarylhawkins : Daryl, please create new issues with a couple of lines of documentation or, better, a link to your comments to the PR, so that we address them at a later time.

wdhawkins (approved):

ragusa:
@zachhardy many times in this PR, some of your answers to comments were to raise an issue to address this later because the comments were made about an implementation choice that was in place in other parts of the code prior to your involvement. I **fully agree with you** that we should not change something here and then not change it over there. However, I am having a hard time keeping track of the different situations where this has happened. Would it be possible for you to create those issues and reference the proper comments in this PR? That would be very helpful moving forward. At this stage, I think PR-154 should be merged but I also think the issues should be reported for a later assignment/discussion. Thanks!

ChiModules/KEigenvalueSolver/IterativeMethods/power_iteration.cc
zachhardy: This issue is addressed in #159 and further brought up in #158.

ChiModules/KEigenvalueSolver/initialize_precursors.cc
zachhardy: This has now been removed.

ChiModules/LinearBoltzmannSolver/lbs_linear_boltzmann_solver.h
zachhardy: This was raised as an issue in #165

ChiModules/KEigenvalueSolver/initialize_precursors.cc
zachhardy: @Naktakala What are your thoughts on this?

zachhardy:
Sure @ragusa .

Below are separate issues/discussions brought up that are outside the scope this PR to be addressed later:

- #165 
- #164 
- #163
- #158  which was a result of a discovered bug addressed in #159 
- #151 
- Removal of temporary memory pointers in various routines discussion.

ragusa:
@zachhardy Thanks!

ChiModules/KEigenvalueSolver/initialize_precursors.cc
Naktakala: Yes, I've come to dislike my style here. Get rid of all the `somethingp` pointers. When we discussed this last week I've looked up the optimizers behavior in this regard. It pretty much optimizes out well.

dominiccaron (approved):

PR #155, opened 2021-06-25
Added ability to run regression tests on class01 on ne cluster.

pbehne:
@Naktakala


Naktakala (approved):

ragusa (approved):
I think ne.tamu.edu will be going away soon. just FYI

Issue #158, opened 2021-06-28
Add more regression suites for reflective BCs


Naktakala:
Regression needs to test LBS' handling of all BCs including opposing reflecting.

PR #159, opened 2021-06-28
Bug fix for reflecting boundary conditions

Naktakala:
## Bug fix for reflecting boundary conditions
Depending on a source flag, boundary sources are activated in a sweep-chunk. The logic in the function that then gets the boundary flux then has to FIRST check if the boundary is reflective before evaluating the boundary source activation. Sometime ago this bug was introduced and influences GMRES with reflective BCs.

### Managerial issue
- We now have additional regression tests that use reflective BCs. The regression suite needs to expand more but this change will be sufficient to fix the bug.
- Issue #158 has been created to address this.


zachhardy (approved):
Move 2 lines of code to fix it and spend 168 lines of code proving that it's fixed. Very nice.

ChiTech/ChiMesh/SweepUtilities/AngleSet/angleset.cc
```diff
@@ -168,13 +168,13 @@ PsiBndry(uint64_t bndry_map,
...
-
   if (ref_boundaries[bndry_map]->IsReflecting())
     return ref_boundaries[bndry_map]->HeterogenousPsiIncoming(
       angle_num, cell_local_id, face_num, fi, gs_ss_begin);
 
+  if (not surface_source_active)
+    return &ref_boundaries[bndry_map]->zero_boundary_flux[g];
+
   return &ref_boundaries[bndry_map]->boundary_flux[g];
 }
```
ragusa, line 180: This is what is happening:
if reflective boundary, return A
if not surf_src_act, return B
return C (all other cases)

what when wrong was that the other was somehow incorrect. the current order makes more sense. however, the way the "logic space" is partitioned still feels weird.  do we test somewhere (input-level) that the suer cannot set a surface source on a face that has a reflective BC?

Naktakala: Regarding the question ... this is essentially covered by the boundary type specification. Each boundary always has the member `zero_boundary_flux` inherited from `BoundaryBase` and all boundaries are by default mapped to a child `BoundaryVacuum` object. When the user calls the `chiLBSSetProperty` routine with the `BOUNDARY_CONDITION` option (and the reflecting option) then boundary-id XX's mapping is changed to a `BoundaryReflecting` object which will return true when `BoundaryBase::IsReflecting()` is called.

It is therefore impossible to have a surface source specified on a reflective boundary. They would be objects of different classes.

I am trying to think how to make the logic-space less weird ... But first I have to understand the cause of the weirdness weighed against the required functionality. Is it maybe because the implementation of boundaries are too far abstracted into `chi_mesh::sweep_utilities`?

dominiccaron (approved):

wdhawkins (approved):

Issue #150, opened 2021-06-17
Include delayed neutrons and improve AGS iteration schemes


Naktakala:
Delayed neutrons are currently not supported in LBS. AGS iteration schemes also need to be improved.

Naktakala:
Addressed in various places

Issue #146, opened 2021-06-17
Transport XS - improve consistency with Chi-prompt versus Chi


Naktakala:
When precursors are used, the KEigen solver uses chi, chi-delayed, nu-prompt and nu-delayed. When precursors are not used it uses chi, and nu. This implies chi is the same between the two. This needs to be corrected.

ragusa:
- ideally, when problem = steady state and nu/chi-prompt and nu/chi-delayed are found, then compute total nu (simple sum) and total-chi (nu-weighted sum of the chi's)
- when problem = steady state and nu/chi-prompt is found but nu/chi-delayed are **not** found, then use nu-prompt as total nu and chi-prompt as total-chi but have a warning message
- when problem = steady state and user entered total nu and chi, then all good.

Many places to check there usage of chi in the code.

Naktakala:
@zachhardy Has this issue been addressed now?

zachhardy:
@Naktakala Yes and no... We modified the `TransportCrossSections` class to distinguish between chi and chi-prompt, but we still need to add checks to ensure the user provides the correct quantities. The `TransportCrossSections` class does not have knowledge of the type of problem being run so these will have to be checks within `LinearBoltzmann::Solver`. Perhaps we could do the following:

- Within `TransportCrossSections`, if nu/chi-prompt/delayed are provided, we can compute the total quantities
- Within `LinearBoltzmann::Solver::PerformInputChecks` we can add checks to ensure the appropriate quantities are available for the type of calculation specified.

Note: This is exactly what @ragusa mentioned in the comment above, just more verbose about implementation.

Naktakala:
I support this. Can you @zachhardy make that happen so we can close this issue?

PR #160, opened 2021-06-29
Cross Section Refurbish

zachhardy:
## Highlights

- Added `chi_prompt` member to `TransportCrossSections`.
- Added various checks.
- Refactored precursor handling in `MakeCombined` routine.
- Modified `SetKSource` routine to reflect `chi_prompt` addition.
- Some renaming.

### `TransportCrossSections` changes

- Added `chi_prompt` member.
- Modified variable names in Lua table to be equivalent to the names in the `TransportCrossSections` class.
- The `MakeCombined` routine now tracks all precursor families across all cross section sets instead of producing an averaged set of precursor families.
    - This was done to allow mingling of fissile materials with different (or no) precursor family structures.
    - `num_precursors` is now equal to the sum of `num_precursors` over the cross section sets being combined.
    - All `precursor_lambda` and `chi_delayed` inputs are stored as written in the ChiTech cross section input files. The `precursor_yield` quantities are normalized based on the relative material density to the total density of all materials with precursors being combined. This is done to preserve a unit sum for yields.
    - Checks are implemented in the reading process to ensure that all spectra within `chi_delayed` sum to one as well as `precursor_yield`. 

### Changes to ChiTech cross-section file syntax

- `DDT_COEFF` changed to `INV_VELOCITY`.
- `PRECURSOR_GAMMA` change to `PRECURSOR_YIELD`.
- `CHI` added to be distinct from `CHI_PROMPT`.

### Comments

- This PR is a first step to address #146 and #150.
- Additional changes to address these issues can be expected after #154 has been approved and merged.


ragusa:
I can't review (not invited) but looks good to me

ragusa (commented):
@Naktakala @zachhardy my only question is what is the default behavior
- for transients? Do we require prompt and delayed values and throw and error is not present?
- for steady-state? Do we require chi and nu_bar (without prompt/delayed suffixes)? Or, if the prompt and delayed ones are present, we compute the nu-weighted values? If an issue needs to be created to do this later, please do so. Thanks!

zachhardy:
@ragusa 
Right now the steady-state solver uses `chi` and `nu` only. I plan to implement delayed neutron capabilities into this once the PRs get accepted and merged. Once this is done `chi` and `nu` will only be used when precursors are off and otherwise the appropriate prompt and delayed quantities.

The transient and k-eigenvalue solvers currently assume that `chi = chi_prompt` and uses that for any and all prompt fission. When precursors are off `nu` is used. When they are on, `nu_prompt` and `nu_delayed` are used. My plan is to do the same as I mentioned above: only use `chi` and `nu` when precursors are off and only use prompt and delayed quantities when they are on.

More checks are definitely necessary. There is a lot we could do with input checks that I can start to work on.

Naktakala (changes requested):
Everything checks out. Just needs a rebase to make it mergeable

Naktakala (approved):

Naktakala:
Are we still waiting for something?

ragusa:
I think I left comments some time ago. I'd like for @dominiccaron and/or @wdarylhawkins to take a look. So far, it's been very fruitful to have many pairs of eyes.

wdhawkins (approved):

PR #161, opened 2021-06-29
File path bug fix in WritePVTU

pbehne:
When writing the PVTU file during the exportation of a field function,
each Piece Source included any full or relative path included in the
base_filename variable. Since the pvtu and vtu files reside in the same
folder, this path is incorrect and should be omitted to allow paraview
to locate the vtu files when reading the pvtu. This commit corrects this
issue.

### DIFF
--- a/ChiTech/ChiPhysics/FieldFunction/fieldfunction_exportvtk.cc
+++ b/ChiTech/ChiPhysics/FieldFunction/fieldfunction_exportvtk.cc
@@ -12,7 +12,6 @@ extern ChiMPI& chi_mpi;
 #include <vtkCellType.h>
 
 #include <fstream>
-#include <sstream>
 
 //###################################################################
 /**Exports a field function to VTK format.
@@ -121,10 +120,7 @@ void chi_physics::FieldFunction::WritePVTU(const std::string& base_filename,
     chi_mesh::GetCurrentHandler()->volume_mesher->options.mesh_global;
   
   // Cut off path to base_filename
-  std::string filename_short;
-  std::istringstream iss(base_filename);
-  while (std::getline(iss, filename_short, '/'))
-    continue;
+  std::string filename_short = base_filename.substr(base_filename.find_last_of("/\\")+1);
 
   for (int p=0; p<chi_mpi.process_count; p++)
   {
--- a/ChiTech/ChiPhysics/FieldFunction/fieldfunction_exportvtk.cc
+++ b/ChiTech/ChiPhysics/FieldFunction/fieldfunction_exportvtk.cc
@@ -12,7 +12,7 @@ extern ChiMPI& chi_mpi;
 #include <vtkCellType.h>
 
 #include <fstream>
-#include<sstream>
+#include <sstream>
 
 //###################################################################
 /**Exports a field function to VTK format.
--- a/ChiTech/ChiPhysics/FieldFunction/fieldfunction_exportvtk.cc
+++ b/ChiTech/ChiPhysics/FieldFunction/fieldfunction_exportvtk.cc
@@ -122,7 +122,7 @@ void chi_physics::FieldFunction::WritePVTU(const std::string& base_filename,
   // Cut off path to base_filename
   std::string filename_short = base_filename.substr(base_filename.find_last_of("/\\")+1);
 
-  for (int p=0; p<chi_mpi.process_count; p++)
+  for (int p=0; p<chi_mpi.process_count; ++p)
   {
     if (is_global_mesh and p!=0) continue;
 
--- a/ChiTech/ChiPhysics/FieldFunction/fieldfunction_exportvtk.cc
+++ b/ChiTech/ChiPhysics/FieldFunction/fieldfunction_exportvtk.cc
@@ -12,6 +12,7 @@ extern ChiMPI& chi_mpi;
 #include <vtkCellType.h>
 
 #include <fstream>
+#include<sstream>
 
 //###################################################################
 /**Exports a field function to VTK format.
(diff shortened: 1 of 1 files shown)


ragusa:
@pbehne you need to invite more reviewers. 2 reviews are needed before merging

pbehne:
> @pbehne you need to invite more reviewers. 2 reviews are needed before merging

I don't think I'm currently allowed to invite reviewers. @Naktakala? Also tagging @rodriguez-jay since he is listed as author of the function.

ragusa:
@pbehne  I invited myself and Jan

ChiTech/ChiPhysics/FieldFunction/fieldfunction_exportvtk.cc
```diff
@@ -12,6 +12,7 @@ extern ChiMPI& chi_mpi;
 #include <vtkCellType.h>
 
 #include <fstream>
+#include<sstream>
```
ragusa: add space for consistency

ChiTech/ChiPhysics/FieldFunction/fieldfunction_exportvtk.cc
```diff
@@ -118,13 +119,19 @@ void chi_physics::FieldFunction::WritePVTU(const std::string& base_filename,
...
   bool is_global_mesh =
     chi_mesh::GetCurrentHandler()->volume_mesher->options.mesh_global;
+  
+  // Cut off path to base_filename
+  std::string filename_short;
+  std::istringstream iss(base_filename);
+  while (std::getline(iss, filename_short, '/'))
+    continue;
 
   for (int p=0; p<chi_mpi.process_count; p++)
```
ragusa: @Naktakala Don;t you prefer ++p ?

ChiTech/ChiPhysics/FieldFunction/fieldfunction_exportvtk.cc
```diff
@@ -118,13 +119,19 @@ void chi_physics::FieldFunction::WritePVTU(const std::string& base_filename,
 
   bool is_global_mesh =
     chi_mesh::GetCurrentHandler()->volume_mesher->options.mesh_global;
+  
+  // Cut off path to base_filename
+  std::string filename_short;
+  std::istringstream iss(base_filename);
+  while (std::getline(iss, filename_short, '/'))
```
ragusa: are you trying to do this? If so, what's wrong with my one-liner?

std::string filename_short = base_filename.substr(base_filename.find_last_of("/\\") + 1)

I let @Naktakala  decide what he prefers

ChiTech/ChiPhysics/FieldFunction/fieldfunction_exportvtk.cc
pbehne:
```suggestion
#include <sstream>
```

ChiTech/ChiPhysics/FieldFunction/fieldfunction_exportvtk.cc
Naktakala: If I define three levels of style with the following words:
- High importance (always incurs an objection),
- Medium importance (might incur an objection), and
- Low importance (unlikely to incur an objection)

Then `p++` on an integer is of low- to medium importance. The low portion is because the optimizer gets rid of that in an instant because it can detect that the return value is not used. The medium portion is because I would like us to be consistent and see all the loops use `++p` instead of `p++`.

So in this case I would say there is no objection, just a friendly "Can you please change that `p++` to `++p`? Pretty pleasseeee!

ChiTech/ChiPhysics/FieldFunction/fieldfunction_exportvtk.cc
Naktakala: The one liner needs to escape the "\\" sybol. I just godbolted both versions:

### PR Version:
```c++
std::string song = "Maya_hi/Maya_hu/Maya_ho/Maya_haha/dragos_din_tei.old";

std::string filename_short;
std::istringstream iss(song);
while (std::getline(iss, filename_short, '/'))

std::cout << filename_short;
```
```
>>Maya_hiMaya_huMaya_hoMaya_hahadragos_din_tei.old
```

### Ragusa version
```c++
std::string song = "Maya_hi/Maya_hu/Maya_ho/Maya_haha/dragos_din_tei.old";

std::string filename_short = song.substr(song.find_last_of("/\\")+1);

std::cout << filename_short;
```

```
>>dragos_din_tei.old
```

So please change to second version.

Naktakala (dismissed):

ChiTech/ChiPhysics/FieldFunction/fieldfunction_exportvtk.cc
ragusa: @Naktakala I am still useful !!

pbehne: @Naktakala, you get a bunch of junk in the PR version because you left out the continue, allowing the cout to execute with each iteration. After the loop, the variable filename_short holds the desired value. I agree that the one-liner is more concise, however.

ragusa (approved):

ChiTech/ChiPhysics/FieldFunction/fieldfunction_exportvtk.cc
Naktakala: @pbehne My bad.

Naktakala (approved):

dominiccaron:
Delayed question: are we precluded from using directly the class ```vtkXMLPUnstructuredGridWriter```? If not, I think that it would be beneficial to use the methods available in the VTK API. It avoids the requirement of maintaining a method that otherwise must be kept coherent with the developments occurring in another project.

Issue #166, opened 2021-06-30
Enter dark-side of C++. No more `for (int`.


Naktakala:
Based on discussion #151, the clang-tidy and/or gcc-equivalent flags `-Wsign-conversion` and `-Wsign-compare` needs to be enforced and the associated warnings removed. This will require a large scale refactor.

Naktakala:
Using the clang-tidy warning `-Wsign-conversion` leads to very yucky warnings.

![image](https://user-images.githubusercontent.com/25042529/125506151-8ba48663-16a2-45ce-ad18-fda3c7218075.png)

I can see now why it is not enabled by default.

Naktakala:
Been trying my best at this. Will improve as time goes by. No longer actively hunting for this.

PR #167, opened 2021-06-30
Group-set capabilities in k-eigenvalue solver + delayed neutron option for steady-state

zachhardy:
## Highlights

- This PR uses functionality from #154 and #160.
- Incorporates group-sets into the `PowerIteration` routine.
- Adds delayed neutron precursor option to steady-state calculations within `LinearBoltzmann::Solver`
- Removes `SetKSource` routine and normalizes the precomputed fission source external to the `SetSource` routine.
- Modifies `Groupset` object by storing a reference to psi vector instead of the vector itself.
- Cleans up some formatting in various places

### Changes to `Groupset`

- `psi_new_local` is now a reference to a vector.
    - Constructor changed to take a vector as an argument.
    - The vector this references is instantiated in the Lua routine that creates the group-set.
    - The vector is stored within a new member of the `LinearBoltzmann::Solver` class, a vector of vectors of doubles called `groupset_psi_new_local`.
    - This change does not impact any areas of the code after instantiation of a group-set and makes the `Groupset` class more lightweight.
    - This change was made in anticipation of a group-set-wise `psi_prev_local` being needed for transient calculations.

### Changes to `LinearBoltzmann::Solver`

- New member `groupset_psi_new_local` added. This is explained in detail above.
- Delayed neutron capabilities added to the `SetSource` routine fission terms.
- Removed temporary memory pointers in `SetSource` routine.

### Changes to `KEigenvalue::Solver`

- Added loop over group-sets within each `PowerIteration`
- Removed member `phi_prev_local`. Precomputing the fission source makes this an unnecessary vector. This could perhaps be added in again later if convergence of the flux profile is desired.
- Various renaming and cleanups.


ChiModules/KEigenvalueSolver/IterativeMethods/power_iteration.cc
```diff
@@ -18,168 +18,127 @@ typedef sweep_namespace::SchedulingAlgorithm SchedulingAlgorithm;
...
+              false);
       }
-    }//for source iterations
 
-    if (!si_converged) {
-      chi_log.Log(LOG_ALLVERBOSE_1)
-          << "\n!!!WARNING!!! "
-             "Source iterations did not converge.\n";
-    }
+      CleanUpWGDSA(groupset);
```
ragusa: shouldn't the clean up be conditional to whether these accel methods were used?

ChiModules/KEigenvalueSolver/k_eigenvalue_initialize.cc
```diff
@@ -0,0 +1,42 @@
...
+    //accumulate precursor count and set max
+    for (auto& xs : material_xs)
+    {
+      num_precursors += xs->num_precursors;
+      if (xs->num_precursors > max_num_precursors_per_material)
+        max_num_precursors_per_material = xs->num_precursors;
+    }
+
+
+    if (num_precursors > 0)
```
ragusa: not sure I fully follow. How can we be inside the conditional use_precursors and then later check if n_prec>0. I think I am missing something.

ChiModules/LinearBoltzmannSolver/lbs_02_main_execute.cc
```diff
@@ -60,19 +60,25 @@ void LinearBoltzmann::Solver::SolveGroupset(LBSGroupset& groupset,
...
+  q_moments_local.assign(q_moments_local.size(), 0.0);
+
   if (groupset.iterative_method == IterativeMethod::CLASSICRICHARDSON)
   {
     ClassicRichardson(groupset, group_set_num, sweep_scheduler,
                       APPLY_MATERIAL_SOURCE |
                       APPLY_AGS_SCATTER_SOURCE |
                       APPLY_WGS_SCATTER_SOURCE |
-                      APPLY_FISSION_SOURCE);
+                      APPLY_AGS_FISSION_SOURCE |
```
ragusa: @wdarylhawkins Do you have a comment on this, from your PDT experience?

ChiModules/LinearBoltzmannSolver/lbs_structs.h
```diff
@@ -41,10 +41,14 @@ struct Options
 
   int max_iterations = 1000;
   double tolerance    = 1e-8;
+
   bool use_precursors = false;
 
   bool save_angular_flux = false;
 
+  bool verbose_inner_iterations = true;
+  bool verbose_outer_iterations = true;
```
ragusa: I suppose all of these parameters can be changed by the user, correct? These are only default values, right?

ChiTech/ChiPhysics/PhysicsMaterial/transportxsections/prop_transportxs_00_general.cc
```diff
@@ -129,71 +133,52 @@ void chi_physics::TransportCrossSections::
...
+      exit(EXIT_FAILURE);
     }
-    ++count;
-  }//for auto
+    chi_log.Log(LOG_0) << "NUMBER OF PRECURSORS" << xs->num_precursors;
+    num_precursors_J += xs->num_precursors;
+  }//for cross-section
+
+  //this should be unnecessary with the added condition check later
   if (Nf_total < 1.0e-28)
```
ragusa: the 1e-28 makes me feel there is a cm^3 units somewhere rather than a more common cm-barn. Can anyone comment on this? Are there assumed units in the code? @Naktakala @dominiccaron @wdarylhawkins

ChiModules/KEigenvalueSolver/k_eigenvalue_initialize.cc
zachhardy: @ragusa Yes this was faulty logic on my part. The thinking was to only do anything regarding precursors if the option is turned on and then to only initialize the vector if there are precursors, but that’s not ideal. I think I will modify this to compute the total number of precursors and max per material outside of any conditional, then, if num_precursors == 0, set use_precursors = false, then initialize the vector is use_precursors == true. I am not at my computer today because I am moving so I will get this done in a few days.

ChiModules/LinearBoltzmannSolver/lbs_structs.h
zachhardy: @ragusa That is correct. There are Lua routines that allow the user to change these in a Lua input file.

ChiModules/KEigenvalueSolver/IterativeMethods/power_iteration.cc
```diff
@@ -18,168 +18,127 @@ typedef sweep_namespace::SchedulingAlgorithm SchedulingAlgorithm;
...
-    chi_log.Log(LOG_0VERBOSE_2)
-      << "\n********** Starting source iterations";
-
-
-    // ----- Start inner source iterations
-    double pw_change_prev = 1.0;
-    bool si_converged = false;
-    for (int si_nit=0; si_nit<groupset.max_iterations; si_nit++)
+    //============================================= Loop over groupsets
+    MPI_Barrier(MPI_COMM_WORLD);
```
wdhawkins: What was the rationale behind this barrier?  I don't see a need for this here (but I might be missing something not included in the PR).  We need to be very careful with barrier placements.

ChiModules/KEigenvalueSolver/IterativeMethods/power_iteration.cc
```diff
@@ -18,168 +18,127 @@ typedef sweep_namespace::SchedulingAlgorithm SchedulingAlgorithm;
...
-      if (si_converged)
-        si_iter_info << " CONVERGED\n";
-      chi_log.Log(LOG_0VERBOSE_1) << si_iter_info.str();
-
-      if (si_converged) {
-        chi_log.Log(LOG_0VERBOSE_1)
-            << "\nSource iterations converged in "
-            << si_nit << " iteratrions.\n";
-        break;
+      ComputeSweepOrderings(groupset);
```
wdhawkins: Can we not store sweep orderings?  If not, we should probably create an issue.  Storage of the orderings should be very lightweight compared to the computation.

ChiModules/KEigenvalueSolver/IterativeMethods/power_iteration.cc
```diff
@@ -18,168 +18,127 @@ typedef sweep_namespace::SchedulingAlgorithm SchedulingAlgorithm;
...
-          << "\n!!!WARNING!!! "
-             "Source iterations did not converge.\n";
-    }
+      CleanUpWGDSA(groupset);
+      CleanUpTGDSA(groupset);
 
-    // ----- Recompute eigenvalue
+      ResetSweepOrderings(groupset);
+
+      MPI_Barrier(MPI_COMM_WORLD);
```
wdhawkins: Is this barrier necessary?  If so, why?

ChiModules/KEigenvalueSolver/initialize_precursors.cc
```diff
@@ -15,45 +15,53 @@ using namespace LinearBoltzmann;
...
+
+      auto xs = material_xs[xs_id];
+
+      //======================================== Loop over nodes
+      const int num_nodes = full_cell_view.NumNodes();
+      for (int i = 0; i < num_nodes; ++i)
+      {
+        size_t ir = full_cell_view.MapDOF(i, 0, 0);
+        size_t jr = discretization->MapDOFLocal(cell, i, precursor_uk_man, 0, 0);
+        double* Nj_newp = &precursor_new_local[jr];
```
wdhawkins: This should be removed and an offset used below.

ChiModules/KEigenvalueSolver/k_eigenvalue_execute.cc
```diff
@@ -0,0 +1,23 @@
...
+
+#include <iomanip>
+
+
+using namespace LinearBoltzmann;
+
+//###################################################################
+void KEigenvalue::Solver::ExecuteKSolver()
+{
+  MPI_Barrier(MPI_COMM_WORLD);
```
wdhawkins: Again, I don't think this barrier or the one below are needed.

ChiModules/LinearBoltzmannSolver/lbs_03f_resetswporderings.cc
```diff
@@ -50,11 +51,12 @@ void LinearBoltzmann::Solver::ResetSweepOrderings(LBSGroupset& groupset)
...
-    << "\n" << std::setprecision(3)
-    << "           Total application memory (max): "
-    << total_app_memory/1000.0 << " GB\n"
-    << "           Maximum process memory        : "
-    << max_proc_memory/1000.0 << " GB\n\n";
+  if (options.verbose_inner_iterations)
+    chi_log.Log(LOG_0)
+      << "\n" << std::setprecision(3)
+      << "           Total application memory (max): "
+      << total_app_memory/1000.0 << " GB\n"
```
wdhawkins: Tiny nitpick, but this should be 1024 (and below for the max).

ChiModules/LinearBoltzmannSolver/lbs_structs.h
```diff
@@ -28,7 +28,7 @@ struct Options
   GeometryType geometry_type = GeometryType::NO_GEOMETRY_SET;
   SDMType sd_type = SDMType::UNDEFINED;
   unsigned int scattering_order=1;
-  int  sweep_eager_limit= 32000;;
+  int  sweep_eager_limit= 32000;
```
wdhawkins: I don't think this bit of code was introduced in this PR, but we should document why we chose this value.  In particular, it should probably be a power of two... especially if it is supposed to correspond to the MPI eager limit or indicate a buffer size.

wdhawkins (changes requested):

ChiTech/ChiPhysics/PhysicsMaterial/transportxsections/prop_transportxs_00_general.cc
Naktakala: This one is actually difficult. There are no units involved here. Please help us out here by suggesting a better minimum value that prevents the trigger that makes a cross section fissile.

ChiModules/KEigenvalueSolver/IterativeMethods/power_iteration.cc
Naktakala: Ooooohhh. Please don't remove that one. He is probably mimicking LBS execute function (which will become AGS iteration). 

When things go wrong with graphs/cycles/FLUDS then that barrier has always given me a point of departure. It makes sense to have that synchronization point before starting to loop over group sets. We can maybe remove it later but for now please keep it. It only bothers performance when simulations are really small.

Pretty much just asking to keep my training wheels until I am a little more comfortable.

ChiModules/KEigenvalueSolver/IterativeMethods/power_iteration.cc
Naktakala: Tried this before and it can get quite beefy when lots of angles are involved. Here's what happens:
- Let's say 10,000 cells on a process
- The order of the cells in the topo-sort alone will require 10,000 x 8-byte ids
- Inter-location dependencies, cycles, location-to-location sweep ordering, all take additional overhead space but lets say we have to save an additional 0.5 times the amount of cells in "overhead"
- Let's say 5,000 angles in a groupset

Thats 10,000 x 1.5 x 5,000 x 8 bytes = 600 Mb.

Times 3 groupsets, prestored? 1.8 Gb. Ouch.

wdhawkins: Interesting.  Having just been through all of this with the GPU sweeper, I think we can save a lot of memory here.  At least half of non-overhead, and a pretty good percentage of overhead by smartly picking anglesets to reduce the number of cycles. I'll create an issue to remind us to look at this.

ChiModules/KEigenvalueSolver/IterativeMethods/power_iteration.cc
Naktakala: Please also do not remove this one. Same reasons as above. I have never identified these barriers to take any appreciable amount of time to clear with any simulation other than a really tiny one and even then the time taken is very small because the num_procs is also low. I have yet to be proven wrong.

ChiModules/KEigenvalueSolver/initialize_precursors.cc
Naktakala: Agreed

ChiModules/LinearBoltzmannSolver/lbs_structs.h
Naktakala: @zachhardy Zach can you add just abbreviated comment here? The input documentation for `chiLBSSetProperty` has this documented well but we probably need to add an abbreviated version here and reference the documentation.

ChiModules/KEigenvalueSolver/IterativeMethods/power_iteration.cc
wdhawkins: I can promise that unnecessary synchronization points run amok will eventually bite us.  Our jobs so far have been modest at best. At 1M+ MPI processes (or less depending on the interconnect), performance will suffer greatly.

ChiModules/KEigenvalueSolver/k_eigenvalue_execute.cc
Naktakala: Mmmm interesting. Friendly question: Are the barrier concerns hard objections or just mild concerns? I am asking because we are still going strong in developing the code and when I see barriers that would obviously cause performance concerns then I would also advise them to be "off with his head". Otherwise, why not leave them? In this case especially, I would actually follow some PETSc recommendations and argue that execute is a HUGE step. Wrap it like this:

```c++
void KEigenvalue::Solver::ExecuteKSolver()
{
  chi_log.Log() << "KEigenvalue::Solver execution started\n";
  MPI_Barrier(MPI_COMM_WORLD);

  PowerIteration();

  chi_log.Log() << "KEigenvalue::Solver execution completed\n";
  MPI_Barrier(MPI_COMM_WORLD);
}
```

What I've seen happening is that `execute` completes and then one of the processors are slightly ahead and encounters a bug in field function exporting routines. Then the developer spends endless amounts of time just to discover that the solver is not to blame.

ChiModules/KEigenvalueSolver/IterativeMethods/power_iteration.cc
Naktakala: That's exactly what happens inside them. Saves us from seeing those lines of code here.

ChiModules/LinearBoltzmannSolver/lbs_02_main_execute.cc
Naktakala: @wdhawkins I echo this question. Was there ever a method in PDT that benefited from splitting AGS-fission and WGS-fission? Out of ignorance (and possible future utility) we split it but we have no justification yet.

ChiModules/KEigenvalueSolver/k_eigenvalue_execute.cc
wdhawkins: In general, I think it's poor practice to add barriers without a very compelling reason. Lawrence and I didn't see eye-to-eye on a lot, but, on this point, I think we would be in complete agreement. In this case, I don't see the need for the initial barrier (what operation are we enforcing synchronization for?), and the second barrier is essentially a duplicate barrier (coming almost immediately after the implicit barrier in the all-reduce in PowerIteration). I do understand the need to add barriers for debugging (or as you say training-wheel purposes). In those cases, the reason should be well documented so that we can remove them later.

ChiModules/LinearBoltzmannSolver/lbs_02_main_execute.cc
wdhawkins: To my knowledge, we never did this in PDT, and I can't think of any benefit off the top of my head.

ChiModules/KEigenvalueSolver/IterativeMethods/power_iteration.cc
wdhawkins: It just dawned on me that you say 5000 angles in a groupset.  Unless I misunderstand, that number should be much, much lower unless we're using single angle aggregation (which wouldn't make much sense since we included considerable overhead for cycles).  Instead of 5000, shouldn't that number be on the order of 8 to a few hundred?

wdhawkins: Poor @zachhardy  got caught up in a design/architecture war here.  :)  To keep from holding him up,  I'm going to create separate issues for this and the MPI barrier work.

ChiModules/KEigenvalueSolver/IterativeOperations/compute_production.cc
```diff
@@ -14,49 +14,59 @@ using namespace LinearBoltzmann;
 /**Compute the total fission production in the problem.*/
 double KEigenvalue::Solver::ComputeProduction()
 {
+  typedef SpatialDiscretization_FE  FE;
+  const auto grid_fe_view = std::static_pointer_cast<FE>(discretization);
```
dominiccaron: Please either implement type-checking before the downcast or use ```std::dynamic_pointer_cast```.

ChiModules/KEigenvalueSolver/initialize_precursors.cc
zachhardy: Fixed.

ChiModules/LinearBoltzmannSolver/IterativeOperations/lbs_setsource.cc
```diff
@@ -11,12 +11,12 @@ extern ChiLog& chi_log;
 //###################################################################
 /**Sets the source moments for the groups in the current group set.
  *
- * \param group_set_num Identifies the groupset under consideration.
- * \param apply_mat_src Flag indicating whether the material source needs
- *        to be applied to the routine. This is useful for GMRES
- *        since the material source only features during the computing b.
- *        On this note we also need to treat inscattering this way.
- * \param suppress_phi_old Flag indicating whether to suppress phi_old.
+ * \param groupset_num Identifies the groupset under consideration.
```
dominiccaron: Thanks for beginning to update this; can you correct also the line: ```  * \param groupset_num Identifies the groupset under consideration.```?

ChiModules/LinearBoltzmannSolver/IterativeOperations/lbs_setsource.cc
```diff
@@ -63,30 +63,28 @@ SetSource(LBSGroupset& groupset,
...
         unsigned int ell = m_to_ell_em_map[m].ell;
 
         size_t ir = full_cell_view.MapDOF(i, m, 0);
-        double* q_mom = &destination_q[ir];
-        double* phi_oldp = &phi_old_local[ir];
 
-        //============================== Loop over groupset groups
-        for (int g = gs_i; g <= gs_f; ++g)
+        //======================================== Loop over groupset groups
+        for (size_t g = gs_i; g <= gs_f; ++g)
```
dominiccaron, line 83: Unfortunately this change creates in an consistency of types among ```g```, ```gs_i``` and ```gs_f```.

ChiTech/ChiPhysics/PhysicsMaterial/transportxsections/prop_transportxs_00_general.cc
dominiccaron: To be dimensionless, make the comparison relative to ```N_total```, for example: ```if (Nf_total <= eps * N_total) ...```, with ```eps``` on the order of machine precision or similar. Also, it is necessary to modify appropriately the fictitious value assigned to ```Nf_total``` to ensure that it is not on the same order of magnitude as any of the constituents.

ChiModules/LinearBoltzmannSolver/IterativeOperations/lbs_setsource.cc
```diff
@@ -124,43 +122,82 @@ SetSource(LBSGroupset& groupset,
...
+          }//if fission_avail
 
           //============================== Apply within-groupset fission
           if (fission_avail and apply_wgs_fission_src)
           {
-            //================================ Loop over groups
+            //============================== Loop over groups
             for (size_t gprime = first_grp; gprime <= last_grp; ++gprime)
             {
               if ((gprime >= gs_i) or (gprime <= gs_f))
```
dominiccaron: Is this possibly an error? Should it not be ```and```?

dominiccaron (changes requested):

ChiTech/ChiPhysics/PhysicsMaterial/transportxsections/prop_transportxs_00_general.cc
zachhardy: Because this PR is based upon #160, this comment is probably best addressed there. I added a check for `is_fissile` and `num_precursors > 0` in the `MakeCombined` routine before dividing by `Nf_total` or `Np_total`. This addressed the concern that initially led to the check for `Nf_total < 1.0e-28`, but does assume that the user supplies non-zero scaling factors for the cross-sections. I would propose something along the lines of this:
```suggestion  
  if (Nf_total < eps * N_total)
    this->is_fissile = false;
  if (Np_total < eps * N_total)
    this->num_precursors = 0;
```
and then later only dividing by `Nf_total` or `Np_total` if `is_fissile == true` or `num_precursors > 0`, respectively.

zachhardy: Because this PR is based upon #160, this comment is probably best addressed there. I added a check for `is_fissile` and `num_precursors > 0` in the `MakeCombined` routine before dividing by `Nf_total` or `Np_total`. This addressed the concern that initially led to the check for `Nf_total < 1.0e-28`, but does assume that the user supplies non-zero scaling factors for the cross-sections. I would propose something along the lines of what is below and then later only dividing by `Nf_total` or `Np_total` if `is_fissile == true` or `num_precursors > 0`, respectively.
```suggestion  
  double eps = std::numeric_limits<double>::epsilon(); //machine precision
  if (Nf_total < eps * N_total)
    this->is_fissile = false;
  if (Np_total < eps * N_total)
    this->num_precursors = 0;
```

zachhardy: This change has been pushed to #160.

ChiModules/LinearBoltzmannSolver/lbs_03f_resetswporderings.cc
zachhardy: Made the change.

ChiModules/LinearBoltzmannSolver/lbs_structs.h
zachhardy: @Naktakala Added a blurb that references the documentation in chiLBSSetProperty. If you want more, this is not my area of expertise for explanation, so perhaps comment what you think is appropriate here.

ChiModules/KEigenvalueSolver/IterativeOperations/compute_production.cc
zachhardy: Done.

ChiModules/LinearBoltzmannSolver/IterativeOperations/lbs_setsource.cc
zachhardy: This is true. This inconsistency also exists between `gprime`, `first_grp`, `last_grp`, `gs_i`, and `gs_f` as well. There are a few ways to address this. This discussion fits under the scope of #166 . For the meantime, I am going to cast `gs_i`, `gs_f`, `first_grp`, and `last_grp` to `size_t` until a more permanent solution is implemented.

ChiModules/LinearBoltzmannSolver/IterativeOperations/lbs_setsource.cc
zachhardy: Done.

ChiModules/LinearBoltzmannSolver/IterativeOperations/lbs_setsource.cc
zachhardy: Great catch. Fixed.

ChiModules/LinearBoltzmannSolver/IterativeOperations/lbs_setsource.cc
```diff
@@ -171,15 +171,15 @@ SetSource(LBSGroupset& groupset,
             //============================== Loop over groups
             for (size_t gprime = first_grp; gprime <= last_grp; ++gprime)
             {
-              if ((gprime >= gs_i) or (gprime <= gs_f))
+              if ((gprime >= gs_i) and (gprime <= gs_f))
```
ragusa, line 174: @Naktakala @dominiccaron @wdarylhawkins @zachhardy Correct me if what I am saying here is wrong. If we do not have enough coverage through tests to catch this, we should reconsider the pace how which we are doing PRs. Shouldn't this have been caught by a test? We need more tests with our PRs, no?

dominiccaron: Valid point: the implementation of new functionalities should be accompanied by corresponding regression tests (of course, there may be existing functionalities in the code base that require regression tests as well). In this particular instance, the existing mono-energetic test of the eigenvalue solver is unable to detect a multi-group issue or a group-set issue, so a new test of the eigenvalue solver with multi-group and group-sets truly is merited. Likely, Zach has an input readily available for such a test.

Naktakala (approved):

ChiModules/LinearBoltzmannSolver/IterativeOperations/lbs_setsource.cc
Naktakala: This is a verification test problem on a new feature not a regression test problem. It basically means that little to no verification test has been applied to multigroup problems. We can slow down PRs all we like but we can only walk by the light that shines at our feet. Additionally, if you slow down PRs then they tend to grow and diverge from the code base.

There was some value in saying that the keigensolver was a core module of ChiTech (mostly to maintain a sort of open-source status). I am making the motion now to freeze it as a little code demo and separate this module until the time that we can supply verification suites.

ragusa:
I can provide 172g XS for U5 and a short python script that gets you the kinf and the flux ratios for an infinite homog problem. A chi-tech regression example would run this with S2, in a cube (or a square or both) with reflections BC on all faces and various options for group sets. This should not be too complicated to add.

This also does not change my opinion that more tests should be provided with PRs, especially when new developments/features are the main purpose of the PR. It is just good practice.

I am I am using email but I have little time so hitting the reply button is faster for me right now.

Naktakala:
> I can provide 172g XS for U5 and a short python script that gets you the kinf and the flux ratios for an infinite homog problem. A chi-tech regression example would run this with S2, in a cube (or a square or both) with reflections BC on all faces and various options for group sets. This should not be too complicated to add. This also does not change my opinion that more tests should be provided with PRs, especially when new developments/features are the main purpose of the PR. It is just good practice. I am I am using email but I have little time so hitting the reply button is faster for me right now.

I absolutely agree with having tests for new features during PRs. My opinion is that the pace of PRs are not necessarily related to identifying tests to run. The fission-stuff in SetSource was split into WGS-fission and AGS-fission in another PR. Remember how we went on a tangent on what the flags mean? (and asking if anybody knew if this is necessary). I don't see how taking longer with that PR would've identified that bug. Some bugs only reveal themselves when you run the specific configuration. In this case the solution is an: Create a regression test for steady-state problems that have fission and mutliple groupsets #175. This in turns spawns issue: Create AGS solver routines instead of having to run Execute multiple times #176.

ragusa:
Simple thoughts:

  *   new features in PR --> PR must include new test
  *   bug identified in PR and no test present that would have caught it --> PR must include test, especially if not corner case.

zachhardy:
Due to the messiness and size of this PR, I will be removing it in favor of a more limited scope one which focuses on delayed neutron capabilities in the steady-state solver and accompanying design changes such as the removal of temporary memory pointers and some type inconsistencies. I will introduce this PR once #160 has been merged.

PR #168, opened 2021-07-01
Compiler flags

dominiccaron:
Preliminary work on the removal of compiler flags that suppress warnings as well as the inclusion of additional compiler flags to enable further warnings. Namely:
- inclusion of the ```-pedantic``` flag for all recognised compilers;
- removal of the ```-Wno-unused-but-set-variable``` flag for GCC (absent for Clang);
- modifications to eliminate warnings triggered by the above operations, as well as existing warnings.

The modifications in this pull request do not attempt to address the issues of:
- warnings related to sign conversion / sign comparison (see #166);
- warnings related to unused variables (the compiler flag ```-Wno-unused-variable``` has been left untouched).

With the modifications identified above, a clean compilation is achieved with recent versions of Clang and GCC, with the exception of some warnings originating from two macro definitions in the VTK API that are being called from a class that appears to have been borrowed from ParaView (```vtkCleanUnstructuredGrid```). For example, GCC reports:
```
In file included from chi-tech/ChiTech/ChiMesh/UnpartitionedMesh/vtkCleanUnstructuredGrid.cc:15:
chi-tech/ChiTech/ChiMesh/UnpartitionedMesh/vtkCleanUnstructuredGrid.h:44:71: warning: extra ‘;’ [-Wpedantic]
   44 |   vtkTypeMacro(vtkCleanUnstructuredGrid, vtkUnstructuredGridAlgorithm);
      |                                                                       ^
      |                                                                       -
chi-tech/ChiTech/ChiMesh/UnpartitionedMesh/vtkCleanUnstructuredGrid.cc:33:46: warning: extra ‘;’ [-Wpedantic]
   33 | vtkStandardNewMacro(vtkCleanUnstructuredGrid);
      |                                              ^
chi-tech/ChiTech/ChiMesh/UnpartitionedMesh/vtkCleanUnstructuredGrid.cc:34:84: warning: extra ‘;’ [-Wpedantic]
   34 | vtkCxxSetObjectMacro(vtkCleanUnstructuredGrid, Locator, vtkIncrementalPointLocator);
```
I observe that this error is present both when linking against VTK v8.2.0 and when linking against VTK v9.0.0. However, I do not propose to make modifications in the files ```vtkCleanUnstructuredGrid.h``` and ```vtkCleanUnstructuredGrid.cc```.


dominiccaron (commented):
These modifications were alluded to in discussion #151.

ChiTech/ChiPhysics/FieldFunction/lua/physics_exportfieldfunc.cc
```diff
@@ -126,13 +126,16 @@ int chiExportMultiFieldFunctionToVTKG(lua_State *L)
...
+      << "Invalid field function handle in chiExportMultiFieldFunctionToVTKG";
     exit(EXIT_FAILURE);
   }
 
 //  ff->ExportToVTKG(base_name,field_name);
 
 //  ff->ExportMultiToVTKG(ff_slave,base_name,field_name);
+
+  throw std::logic_error("chiExportMultiFieldFunctionToVTKG : not implemented.");
+
```
dominiccaron: The choice on how to proceed in this case may be related to the upcoming refactor of the field functions mentioned in pull request #152. However, if it is known at this time that there is no intention to implement the underlying C++ functionality upon which the Lua wrapper is based, then I propose to remove the Lua wrapper as part of this pull request. In the meantime, I have added the exception as a precaution.

ChiTech/ChiMesh/Cell/cell.h
```diff
@@ -65,13 +65,10 @@ class Cell
...
   const CellType cell_type;
-  const CellType cell_sub_type;
 
 public:
-  explicit Cell(CellType in_cell_type,
-                CellType in_cell_sub_type=CellType::HEXAHEDRON) :
-                cell_type(in_cell_type),
-                cell_sub_type(in_cell_sub_type)
+  explicit Cell(CellType in_cell_type) :
+                cell_type(in_cell_type)
```
Naktakala, line 71: Mmmm. This still needs to be implemented. Created issue #169

ChiTech/ChiPhysics/FieldFunction/lua/physics_exportfieldfunc.cc
Naktakala: This particular lua function needs to be removed. I don't know what it is doing there. Might have lingered there during development. Can you please just remove it?

Naktakala (changes requested):

ChiTech/ChiPhysics/FieldFunction/lua/physics_exportfieldfunc.cc
dominiccaron: Removed.

wdhawkins (approved):

CMakeLists.txt
```diff
@@ -119,14 +119,17 @@ set(CHI_LIBS stdc++ lua m dl ${MPI_CXX_LIBRARIES} petsc ${VTK_LIBRARIES})
...
     set(CMAKE_CXX_FLAGS_DEBUG  "${CMAKE_CXX_FLAGS_DEBUG} -O0")
 elseif(CMAKE_CXX_COMPILER_ID STREQUAL "AppleClang")
-    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall -Wno-unused-variable -Wno-c++17-extensions")
+    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -pedantic")
+    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall -Wno-unused-variable")
     set(CMAKE_CXX_FLAGS_DEBUG  "${CMAKE_CXX_FLAGS_DEBUG} -O0")
 elseif("${CMAKE_CXX_COMPILER_ID}" STREQUAL "GNU")
-    set(CMAKE_CXX_FLAGS  "${CMAKE_CXX_FLAGS} -Wall -Wno-unused-variable")
-    set(CMAKE_CXX_FLAGS  "${CMAKE_CXX_FLAGS} -Wno-unused-but-set-variable -Wno-sign-compare")
+    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -pedantic")
```
Naktakala, line 130: We need to find a way to not involve pedantic checks for external libraries. With gcc10.2 the compile warnings are going crazy with things like

```
VTK-8.2.0/Common/DataModel/vtkPointData.h:36:50: warning: extra ';' [-Wpedantic]
```

Can we turn it off somehow for certain includes?

Naktakala (changes requested):

CMakeLists.txt
wdhawkins: Yes, I really hate this.  The best way is probably to use -isystem for some includes.  If we have issues with that, we can use individual pragmas, but that's likely to get messy fast.  -isystem should work for gcc and clang.

zachhardy: @Naktakala With GCC-11.1.0 and VTK-9.0.1 I get these same warnings.

ChiModules/DiffusionSolver/chi_diffusion_utils.cc
```diff
@@ -25,11 +25,13 @@ KSPMonitorAChiTech(KSP ksp, PetscInt n, PetscReal rnorm, void *monitordestroy)
     char buff[100];
     if (rnorm/rhs_norm < 1.0e-2)
     {
-      snprintf(buff,100,"Diffusion iteration %4lld - Residual %.3e\n",n,rnorm/rhs_norm);
+      snprintf(buff,100,"Diffusion iteration %4ld - Residual %.3e\n",
+               n,rnorm/rhs_norm);
```
zachhardy: This likely doesn't matter much, but I get compiler warnings such as:

`warning: format '%ld' expects argument of type 'long int', but argument 4 has type 'PetscInt' {aka 'long long int'} [-Wformat=]`

with GCC-11.1.0 and PETSc-3.15.1

ChiTech/ChiMath/PETScUtils/petsc_utils_03_createKSP.cc
```diff
@@ -119,12 +119,12 @@ PetscErrorCode chi_math::PETScUtils::GeneralKSPMonitor(
   KSPGetOptionsPrefix(ksp,&ksp_name);
   if (rnorm/rhs_norm < 1.0e-2)
   {
-    snprintf(buff,100,"%s iteration %4lld - Residual %.3e\n",
+    snprintf(buff,100,"%s iteration %4ld - Residual %.3e\n",
```
zachhardy: Echo the comment on `ChiModules/DiffusionSolver/chi_diffusion_utils.cc`

ChiTech/ChiMesh/MeshContinuum/chi_meshcontinuum_exportpython.cc
```diff
@@ -123,7 +123,7 @@ ExportCellsToPython(const char* fileName, bool surface_only,
 
       for (uint64_t v=0; v<poly_cell.vertex_ids.size(); v++)
       {
-        fprintf(of,"face_vertindi[%d][%llu]=%llu\n",cell_g_index,v,poly_cell.vertex_ids[v]);
+        fprintf(of,"face_vertindi[%d][%lu]=%lu\n",cell_g_index,v,poly_cell.vertex_ids[v]);
```
zachhardy: This likely doesn't matter much, but I get compiler warnings such as:

`warning: format '%lu' expects argument of type 'long unsigned int', but argument 4 has type 'uint64_t' {aka 'long long unsigned int'} [-Wformat=]`

using GCC-11.1.0

zachhardy (commented):
Because these things are not my area of expertise, I will let others approve of this.

Overall, I get the same warnings that were brought up in the PR description and some additional warnings about type incompatibilities explained within the comments of the code in this review. I am using GCC-11.1.0, VTK-9.0.1, and PETSc-3.15.1. I did not add a comment to each instance of the type incompatibility warnings, but all of the changes of the same kind within the PR gave the respective warnings I provided in the review.

CMakeLists.txt
dominiccaron: @Naktakala, @zachhardy: Based on the behaviour that you describe, it appears that, in your build environments, the VTK includes are not correctly registered as system headers, which ultimately are exempt from compiler warnings through the mechanism described by @wdhawkins. Ideally, that occurs internally in ```FindVTK``` and it should not necessary for us to reclassify the include directories produced by the execution of the macro. Of course, we can implement a correction if it turns out to be necessary. However, we need to understand the underlying reason(s) for which this behaviour occurs in your build environments, and why it does not occur in mine, in order for me to be able to make an appropriate modification. Otherwise, one of you will need to undertake the task as I am unable to reproduce the behaviour described by you in any of my multiple build environments. Further coordination to follow offline.

ChiModules/DiffusionSolver/chi_diffusion_utils.cc
dominiccaron: I observed the opposite issue when compiling on various different machines. Anyway, I have decided to transition these parts of the code to C++-style streams, which should avoid system-/implementation-dependent issues altogether.

CMakeLists.txt
wdhawkins: Here's my guess as what's going on.  By default, gcc (and probably clang) treats headers found in certain directories as system headers.  I haven't had time to look up the list, but I suspect it's locations such as /usr/include, /include, etc. This list is configurable during gcc compilation.  On machines where the headers aren't in system directories and a default gcc install is being used (class01 comes to mind... since I couldn't convince IT to let us manage it ourselves), we'll get these warnings without using -isystem.  This probably also applies to installations on laptops, etc. where prerequisite packages were installed in non-system locations for various reasons.

Naktakala: @wdhawkins I think you are right Daryl. `-isystem` seems to have no effect on my laptop.

ChiTech/ChiMath/PETScUtils/petsc_utils_03_createKSP.cc
```diff
@@ -114,23 +116,19 @@ PetscErrorCode chi_math::PETScUtils::GeneralKSPMonitor(
...
-  chi_log.Log(LOG_0) << buff;
 
+  std::stringstream buff;
+  buff
+    << ksp_name
+    << " iteration "
+    << std::setw(4) << n
+    << " - Residual "
+    << std::scientific << std::setprecision(7) << rnorm / rhs_norm
+    << std::endl;
```
dominiccaron, line 129: I removed the conditional that controlled the output formatting. Hopefully there are no objections.

CMakeLists.txt
dominiccaron: In general, I agree with the assessment. However, that situation describes exactly one of my build environments (system compiler, non-system dependencies) in which the configuration occurs correctly. In that case, the ```-isystem``` flag appears correctly in the compiler commands, which is fundamental to the suppression of the warnings from VTK. Perhaps we are not using ```FindVTK``` in the most robust manner, as something appears to contribute to a change of behaviour in our differing systems and/or environments? Anyway, this is good, as it assists to focus our attention on the input to/output of that macro.

Naktakala (approved):

PR #172, opened 2021-07-03
Test runner - more thorough checking for cluster

dominiccaron:
The logic presently employed to determine whether or not the test runner is executed on the NUEN cluster returns a positive result when executed on any NUEN departmental machine for which the hostname contains ```ne.tamu.edu```. This modification provides a more comprehensive examination of the hostname.


ragusa (approved):
@dominiccaron soon, the ne. will disappear. It may be replaced by engr. Just be on the lookout for that.

Naktakala (approved):

PR #178, opened 2021-07-11
Fix few places where the wrong pointer was being checked

jthano:
This is a small bug fix. See the few code changes for details.

### DIFF
--- a/ChiModules/LinearBoltzmannSolver/lua/lbs_compute_balance.cc
+++ b/ChiModules/LinearBoltzmannSolver/lua/lbs_compute_balance.cc
@@ -21,7 +21,8 @@ int chiLBSComputeBalance(lua_State *L)
 
   //============================================= Get pointer to solver
   int solver_handle = lua_tonumber(L, 1);
-  auto lbs_solver = LinearBoltzmann::lua_utils::GetSolverByHandle(solver_handle, "chiLBSComputeBalance");
+  auto lbs_solver = LinearBoltzmann::lua_utils::
+    GetSolverByHandle(solver_handle, __FUNCTION__);
 
   lbs_solver->ComputeBalance();
 
--- a/ChiModules/LinearBoltzmannSolver/lua/lbs_compute_balance.cc
+++ b/ChiModules/LinearBoltzmannSolver/lua/lbs_compute_balance.cc
@@ -27,12 +27,10 @@ int chiLBSComputeBalance(lua_State *L)
   int solver_index = lua_tonumber(L,1);
 
   //============================================= Get pointer to solver
-  chi_physics::Solver* psolver;
   LinearBoltzmann::Solver* solver;
   try{
-    psolver = chi_physics_handler.solver_stack.at(solver_index);
 
-    solver = dynamic_cast<LinearBoltzmann::Solver*>(psolver);
+    solver = dynamic_cast<LinearBoltzmann::Solver*>(chi_physics_handler.solver_stack.at(solver_index));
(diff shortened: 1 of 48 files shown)


ChiModules/LinearBoltzmannSolver/lua/lbs_compute_balance.cc
```diff
@@ -34,7 +34,7 @@ int chiLBSComputeBalance(lua_State *L)
 
     solver = dynamic_cast<LinearBoltzmann::Solver*>(psolver);
 
-    if (not psolver)
+    if (not solver)
```
wdhawkins: Why is psolver even defined?  Is it ever used except in the cast?

jthano: Not used anymore, `psolver` was used as part of an if-statement in a much older version of chi-tech. I'm fine collapsing the code a bit and getting rid of `LinearBoltzmann::Solver* solver` if that's what we want to do. I could quickly go through chi-tech and make the changes everywhere this code logic is present which includes several places not affected by this bug.

wdhawkins: IMHO, simpler is always better.  I would say get rid of it if it's not needed. It prevents exactly these types of issues.

Naktakala: I like the way you collapsed them now. I have another suggestion/kind-request. The scope is to create (in the `lua` directory of LBS) a header file `lbs_lua_utils.h` and a source file `lbs_lua_utils.cc`. Inside the header add to the `LinearBoltzmann` namespace the namespace `lua_utils` and add the prototype

```c++
namespace LinearBoltzmann
{
  namespace lua_utils
  {
     LinearBoltzmann::Solver* GetSolverByHandle(int handle, const std::string& calling_function_name);
  }
}
```

Then in the .cc file define the function:
```c++
LinearBoltzmann::Solver* LinearBoltzmann::lua_utils::
  GetSolverByHandle(int handle, const std::string& calling_function_name)
{
  LinearBoltzmann::Solver* lbs_solver;
  try{

    lbs_solver = dynamic_cast<LinearBoltzmann::Solver*>(
      chi_physics_handler.solver_stack.at(handle));

    if (not lbs_solver)
      throw std::logic_error(calling_function_name + ": Invalid solver at given handle (" +
                             std::to_string(handle) + "). "
                             "The solver is not of type LinearBoltzmann::Solver.");
  }//try
  catch(const std::out_of_range& o) {
    throw std::logic_error(calling_function_name + ": Invalid solver-handle (" +
                           std::to_string(handle) + ").");
  }
}
```

Then in all lua-functions that need to get the solver by handle replace the long-winded code with

```c++
int solver_handle = lua_tonumber(L, 1);
auto lbs_solver = LinearBoltzmann::lua_utils::GetSolverByHandle(solver_handle, fname);
```

Feel free to say no and we just make an issue for it.

Naktakala: Issue #180 created to eventually rename the namespace `LinearBoltzmann` to `linear_boltzmann`.

Naktakala (approved):

zachhardy (changes requested):
I like the changes. This needs a rebase onto development to get caught up with more recent changes, then it is good to go.

zachhardy (approved):

PR #181, opened 2021-07-15
A few additions to lua utilities

Naktakala:
## Highlights
- Lua input-checking routines got the calling function's text name via a c-style string. Now it gets it as a `std::string`. Example:
```c++
//Was
void LuaPostArgAmountError(const char* func_name,int expected, int given);

//Now
void LuaPostArgAmountError(const std::string& func_name,int expected, int given);
```
- We already had `LuaCheckNilValue` and `LuaCheckTableValue` but now we have also
  - `LuaCheckStringValue`
  - `LuaCheckBoolValue`
  - `LuaCheckNumberValue`
  - `LuaCheckIntegerValue`

- Added a utility function `chiGetFieldFunctionHandleByName` which will more easily be able to handle global field-functions across solvers. This will require some future thinking on the naming of field-functions but for now will be pretty useful.


ChiTech/ChiPhysics/FieldFunction/lua/physics_getfieldfunclist.cc
```diff
@@ -55,3 +55,54 @@ int chiGetFieldFunctionList(lua_State* L)
...
+
+  if (num_handles > 1)
+    chi_log.Log(LOG_0WARNING) << fname << ": A total of " << num_handles
+                              << " field-functions were found that matched the "
+                              << " requested name. Only the first match will be "
+                              << " returned.";
+
+  lua_pushinteger(L, static_cast<lua_Integer>(handles_that_matched.front()));
+  return 1;
+}
```
zachhardy, line 108: As you mentioned, more thought will need to be given to field function naming in the future to prevent this type of behavior that could easily yield an undesired result. Perhaps a check for duplicate named field functions or prepended solver names to field function names would work provided no duplicate solver names are allowed. For now, this works just fine though.

zachhardy (approved):

ChiTech/ChiPhysics/FieldFunction/lua/physics_getfieldfunclist.cc
Naktakala: Yes. Prepending the solver-name is on my thoughts. Still thinking about but I think solver names should become mandatory. And we need to think how to handle duplicate names.

ChiTech/ChiLua/chi_lua.cc
```diff
@@ -1,7 +1,6 @@
 #include "chi_lua.h"
 
 #include "chi_log.h"
```
dominiccaron, line 3: ```#include "chi_log.h"```, ```#include <sstream>```, ```#include <map>``` appear to be unnecessary.

ChiTech/ChiPhysics/FieldFunction/lua/physics_getfieldfunclist.cc
dominiccaron: My opinion: do not allow duplicate field names. Here, the name is used as a key; keys should be unique. If the key is not unique, then the concept of retrieving by key has no meaning, as additional information is required to identify fully the entry to which the key corresponds in the event of multiple occurrences of the non-unique key. Of course, it is possible to provide a mechanism to return either all occurrences or to return the n-th occurrence, but that assumes that the user knows how to interpret the ordering, which may not be case.

dominiccaron (approved):

PR #179, opened 2021-07-15
Delayed capabilities in steady state solver

zachhardy:
### Highlights

- Implemented delayed neutron functionality into `LinearBoltzmann::Solver::SetSource`. This is based off of a substituting the delayed neutron precursor concentration term in terms of the delayed fission rate.
- Created a `psi_new_local` vector of vectors within `LinearBoltzmann::Solver` to keep data within the solver and not in the otherwise lightweight `LBSGroupset`. Now, `LBSGroupset` holds a reference to the appropriate `psi_new_local` vector. 
- General cleanup from suggestions in last (cancelled) PR.
- Moved `KEigenvalueSolver` specific items out of the `Options` struct.

PR #183, opened 2021-07-22
LBS Source moments finally 

Naktakala:
## Highlights
- LBS gear-lever added to use either traditional material/boundary source or
  source-moments supplied from file.
- Routines added to write and read source-moments.
- Some file renaming.
- Some doxygen documentation started which will hold more comprehensive 
  documentation in future. 
  of simulations will go there one day.

## Detail - LinearBoltzmannSolver
- A `Documentation` folder has been added to the main LBS folder. The documents
  in this folder contain basically nothing right now but will be used in future 
  to enhance the online documentation. I.e., how to to a poor-man's uncollided flux pair
  of simulations will go there one day.
- Source moments binary file format in `.data` files defined. Writing and 
  reading this type of file is done via 
  - `LinearBoltzmann::Solver::WriteFluxMoments`, and
  - `LinearBoltzmann::Solver::ReadFluxMoments` 
- An `ext_src_moments_local` vector has been added to the LBS-solver to hold
  the optional source-moments. This vector is not used or allocated unless 
  the appropriate options and routines are called.
  - An LBS option `use_src_moments` has been added to control whether the 
    `ext_src_moments_local` is to be used instead of the material/boundary source.
  - This option can be set via CLI with `chiLBSSetProperty`.
- The `SetSource`-method has been modified to be able to use the source-moments  
- `GMRES` and `ClassicRichardson` modified to appropriately channel 
  material/boundary sources or source-moments into the `SetSource`-method
- Two regression tests added.

## Detail - Small addition to SpatialDiscretization
- There are a number of potential problems with reading source-moments from an
  external file.
  - The file could've been generated with an external code.
  - Cell global-id's could not align
  - Cell node numbering could not align
- The last two items can very effectively addressed by obtaining the 
  node-locations for a specific cell for a given spatial discretization
  - `FiniteVolume` will return 1 node at the cell centroid
  - `PWLD` will return nodes co-located on the cell vertices
  - `HigherOrder` spatial discretizations will return the appropriate node
    locations.
- To this effect we now have `GetCellNodeLocations` on all the spatial 
  discretizations. This can be used to compare/map the node locations defined
  in the `.data` file.


ChiTech/ChiMath/SpatialDiscretization/FiniteElement/PiecewiseLinear/pwl.h
```diff
@@ -141,6 +141,18 @@ class SpatialDiscretization_PWLD : public SpatialDiscretization_FE
...
+  {
+    std::vector<chi_mesh::Vector3> node_locations;
+    node_locations.reserve(cell.vertex_ids.size());
+
+    for (auto& vid : cell.vertex_ids)
+      node_locations.emplace_back(ref_grid->vertices[vid]);
+
+    return node_locations;
+  }
+
```
zachhardy, line 155: `SpatialDiscretization_PWLD` and `SpatialDiscretization_PWLC` `GetCellNodeLocations` methods are identical. I was going to suggest putting this into `SpatialDiscretization_FE` since both classes inherit from this, but then it would be tied to linear elements. When higher order elements are implemented, it would likely be best to have a routine with `SpatialDiscretization_FE` that gives node locations for whatever order finite elements are supported.

ChiTech/ChiMath/SpatialDiscretization/FiniteElement/PiecewiseLinear/pwlc.h
```diff
@@ -126,6 +126,18 @@ class SpatialDiscretization_PWLC : public SpatialDiscretization_FE
...
+  {
+    std::vector<chi_mesh::Vector3> node_locations;
+    node_locations.reserve(cell.vertex_ids.size());
+
+    for (auto& vid : cell.vertex_ids)
+      node_locations.emplace_back(ref_grid->vertices[vid]);
+
+    return node_locations;
+  }
+
```
zachhardy, line 140: Echo my other comment.

ChiModules/LinearBoltzmannSolver/lua/lbs_io_flux_moments.cc
```diff
@@ -0,0 +1,140 @@
...
+                          single_file_flag);
+
+  if (file_is_flux)
+  {
+    chi_log.Log() << "Making source moments from flux file." << single_file_flag << file_is_flux;
+    auto temp_phi = solver->phi_old_local;
+    solver->phi_old_local = solver->ext_src_moments_local;
+    solver->MakeSourceMomentsFromPhi(solver->ext_src_moments_local);
+    solver->phi_old_local.assign(solver->phi_old_local.size(),0.0);
+  }
```
zachhardy: What is the purpose of the creation of `temp_phi`? I could be missing something, but it does not appear to be used anywhere.

zachhardy (commented):
Looks good overall. The glory of Git collaboration is getting to say that a rebase is needed on nearly every PR that comes through.

ChiModules/LinearBoltzmannSolver/lbs_04c_io_flux_moments.cc
```diff
@@ -0,0 +1,326 @@
...
+      << __FUNCTION__ << "Failed to open " << file_name;
+    return;
+  }
+
+  //============================================= Write header
+  std::string header_info =
+    "Chi-Tech LinearBoltzmann: Flux moments file\n"
+    "Header size: 500 bytes\n"
+    "Structure(type-info):\n"
+    "size_t num_local_nodes\n"
```
wdhawkins: My suggestion would be to use a platform-independent binary format for this file.  If are going to stick with writing the file ourselves, we should make it as portable as possible.  All of the variables here should have specified sizes (uint64_t, int64_t, uint16_t, etc.).  We shouldn't be using things like size_t.

ChiModules/LinearBoltzmannSolver/lua/lbs_io_flux_moments.cc
```diff
@@ -0,0 +1,140 @@
...
+  {
+    LuaCheckBoolValue(__FUNCTION__, L, 4);
+    file_is_flux = lua_toboolean(L, 4);
+  }
+
+  //============================================= Get pointer to solver
+  chi_physics::Solver* psolver;
+  LinearBoltzmann::Solver* solver;
+  try{
+    psolver = chi_physics_handler.solver_stack.at(solver_index);
```
wdhawkins: We should simplify this code per https://github.com/chi-tech/chi-tech/pull/178#issue-687402142.

ChiModules/LinearBoltzmannSolver/lbs_04c_io_flux_moments.cc
```diff
@@ -0,0 +1,326 @@
...
+
+  //============================================= Write per dof data
+  try {
+    for (const auto& cell : grid->local_cells)
+      for (unsigned int i=0; i<sdm->GetCellNumNodes(cell); ++i)
+        for (unsigned int m=0; m<num_moments_t; ++m)
+          for (unsigned int g=0; g<num_groups; ++g)
+          {
+            uint64_t dof_map = sdm->MapDOFLocal(cell,i,flux_moments_uk_man,m,g);
+            double value = flux_moments.at(dof_map);
```
wdhawkins: at() is going to be slow, and I will argue should be used sparingly.  Instead, we should refactor this code to use [] with an assertion for bounds checking.

ChiModules/LinearBoltzmannSolver/lua/lbs_io_flux_moments.cc
```diff
@@ -0,0 +1,140 @@
...
+  LuaCheckNilValue(__FUNCTION__,L,2);
+
+  int      solver_index = lua_tonumber(L,1);
+  std::string file_base = lua_tostring(L,2);
+
+  //============================================= Get pointer to solver
+  chi_physics::Solver* psolver;
+  LinearBoltzmann::Solver* solver;
+  try{
+    psolver = chi_physics_handler.solver_stack.at(solver_index);
```
wdhawkins: Simplify per https://github.com/chi-tech/chi-tech/pull/178#issue-687402142.

ChiModules/LinearBoltzmannSolver/lbs_04c_io_flux_moments.cc
```diff
@@ -0,0 +1,326 @@
...
+extern ChiMPI& chi_mpi;
+
+#include <fstream>
+#include <cstring>
+
+//###################################################################
+/**Makes a source-moments vector from scattering and fission based
+ * on the latest phi-solution.*/
+void LinearBoltzmann::Solver::
+  MakeSourceMomentsFromPhi(std::vector<double> &source_moments)
```
wdhawkins: I would suggest having this method return the source moments vector instead of taking as an argument (rvo should make it just as fast).

wdhawkins (changes requested):

ChiModules/LinearBoltzmannSolver/IterativeOperations/lbs_setsource.cc
```diff
@@ -140,29 +142,9 @@ SetSource(LBSGroupset& groupset,
...
-                  for (size_t j = 0; j < xs->num_precursors; ++j)
-                    infission_g += xs->chi_delayed[g][j] *
-                                   xs->precursor_yield[j] *
-                                   xs->nu_delayed_sigma_f[gprime] *
-                                   phi_old_local[ir + gprime];
-                }
-              }
+                infission_g += xs->chi[g] *
+                               xs->nu_sigma_f[gprime] *
+                               phi_old_local[uk_map + gprime];
```
dominiccaron: Is the removal of the distinction between total/delayed the result of a mistake during a rebase?

ChiModules/LinearBoltzmannSolver/IterativeOperations/lbs_setsource.cc
```diff
@@ -172,33 +154,13 @@ SetSource(LBSGroupset& groupset,
...
-                    infission_g += xs->chi_delayed[g][j] *
-                                   xs->precursor_yield[j] *
-                                   xs->nu_delayed_sigma_f[gprime] *
-                                   phi_old_local[ir + gprime];
-                }
-              }
+              if ((gprime >= gs_i) or (gprime <= gs_f))
+                infission_g += xs->chi[g] *
+                               xs->nu_sigma_f[gprime] *
+                               phi_old_local[uk_map + gprime];
```
dominiccaron: As above: possibly a mistake during a rebase?

ChiModules/LinearBoltzmannSolver/lbs_04c_io_flux_moments.cc
```diff
@@ -0,0 +1,326 @@
...
+  //============================================= Clean-up
+  file.close();
+}
+
+
+//###################################################################
+/**Writes a flux-moments vector from a file in the specified vector.*/
+void LinearBoltzmann::Solver::ReadFluxMoments(const std::string &file_base,
+                                              std::vector<double>& flux_moments,
+                                              bool single_file/*=false*/)
```
dominiccaron: Please update documentation ("Writes" -> "Reads", etc.).

ChiModules/LinearBoltzmannSolver/lua/lbs_setproperty.cc
```diff
@@ -402,6 +401,16 @@ int chiLBSSetProperty(lua_State *L)
 
     chi_log.Log() << "LBS option to save angular flux set to " << save_flag;
   }
+  else if (property == USE_SOURCE_MOMENTS)
```
dominiccaron, line 389: Can this option be given a write-up in the large documentation string above?

ChiTest/Z_Run_all.py
```diff
@@ -297,6 +297,24 @@ def run_test(file_name, comment, num_procs,
...
+                              ["[0]  Max-value2=", 0.000000000, 1.0e-10]])
+
+run_test(
+    file_name="Transport3D_1Poly_qmom_part2",
+    comment="3D LinearBSolver Test Ortho Grid Parmetis - PWLD",
+    num_procs=4,
+    search_strings_vals_tols=[["[0]  Max-value1=", 1.08197e-01, 1.0e-6],
+                              ["[0]  Max-value2=", 9.14681e-06, 1.0e-10]])
+
+#----------------------------------------------------
```
dominiccaron: Please update the comments and the control values here (and, if necessary) in the corresponding Lua files.

ChiModules/LinearBoltzmannSolver/lua/lbs_io_flux_moments.cc
```diff
@@ -0,0 +1,140 @@
...
+    LuaCheckBoolValue(__FUNCTION__, L, 3);
+    single_file_flag = lua_toboolean(L, 3);
+  }
+
+  bool file_is_flux = false;
+  if (num_args == 4)
+  {
+    LuaCheckBoolValue(__FUNCTION__, L, 4);
+    file_is_flux = lua_toboolean(L, 4);
+  }
```
dominiccaron: This parameter will benefit from further clarification in the doucmentation. If I interpret correctly, ```file_is_flux``` determines whether the file being read contains moments of the source or moments of the flux, right? If true, then perhaps we should consider different nomenclature, as ```(chi)ReadFluxMoments``` / ```(chi)WriteFluxMoments``` both come across as being specific to the flux.

dominiccaron (changes requested):

ChiModules/LinearBoltzmannSolver/lbs_04c_io_flux_moments.cc
Naktakala: Could you add more detail on exactly what you are suggesting/requesting?

ChiModules/LinearBoltzmannSolver/lbs_04c_io_flux_moments.cc
Naktakala: Ahh, yes. Good one.

ChiModules/LinearBoltzmannSolver/lua/lbs_setproperty.cc
Naktakala: Done

ChiModules/LinearBoltzmannSolver/lbs_04c_io_flux_moments.cc
Naktakala: Done

ChiModules/LinearBoltzmannSolver/lua/lbs_io_flux_moments.cc
Naktakala: Fixed. It was supposed to save phi_old_local so that we can restore it.

ChiModules/LinearBoltzmannSolver/lua/lbs_io_flux_moments.cc
Naktakala: Done

ChiModules/LinearBoltzmannSolver/lua/lbs_io_flux_moments.cc
Naktakala: Done

ChiModules/LinearBoltzmannSolver/lbs_04c_io_flux_moments.cc
Naktakala: Done

ChiModules/LinearBoltzmannSolver/IterativeOperations/lbs_setsource.cc
Naktakala: Good catch. Fixed

ChiModules/LinearBoltzmannSolver/IterativeOperations/lbs_setsource.cc
Naktakala: Yes thank you. Fixed

ChiModules/LinearBoltzmannSolver/lbs_04c_io_flux_moments.cc
Naktakala: Done

ChiTest/Z_Run_all.py
Naktakala: Done

ChiModules/LinearBoltzmannSolver/lua/lbs_io_flux_moments.cc
Naktakala: This has been reworked so that we only have the `single_file_flag`. Other wrappers have been added:
- `chiLBSWriteFluxMoments`
- `chiLBSCreateAndWriteSourceMoments`
- `chiLBSReadFluxMomentsAndMakeSourceMoments`
- `chiLBSReadSourceMoments`
- `chiLBSReadFluxMoments` (Also double for reading initial `phi_old_local`)

ChiModules/LinearBoltzmannSolver/lbs_04c_io_flux_moments.cc
wdhawkins: at() always performs bounds checking... something not usually done in production code.  You rarely see at() used specifically for this reason.  Instead, we should do something along the lines of:
```suggestion
            assert(dof_map < flux_moments.size());
            double value = flux_moments[dof_map];
```
Or, just leave out the assert entirely if we trust the value coming from MapDOFLocal().

ChiModules/LinearBoltzmannSolver/IterativeOperations/lbs_setsource.cc
zachhardy: @Naktakala  @dominiccaron Okay time out, #179 addressed #150. Are we intending to revert those changes now?

ChiModules/LinearBoltzmannSolver/IterativeOperations/lbs_setsource.cc
zachhardy: As above: Why do we think it is a mistake to have changes from a previously merged PR (#179) in this PR?

ChiModules/LinearBoltzmannSolver/IterativeOperations/lbs_setsource.cc
Naktakala: @zachhardy We should not be reverting so please help me determine if the rebase is consistent now. Please check the raw file since this comment-stream is now based on an outdated diff. The only major change in this routine should be replacing
```c++
if (apply_mat_src && (ell == 0))
  destination_q[ir + g] += src[g];
```
with
```c++
if ( apply_mat_src and (ell == 0) and (not options.use_src_moments))
  destination_q[uk_map + g] += src[g];
else if (apply_mat_src and options.use_src_moments)
  destination_q[uk_map + g] += ext_src_moments_local[uk_map + g];
```
and of course the renaming of `ir` to `uk_map`.

ChiModules/LinearBoltzmannSolver/IterativeOperations/lbs_setsource.cc
Naktakala: Echo above. Please check the updated diff

ChiTech/ChiMath/SpatialDiscretization/FiniteElement/PiecewiseLinear/pwl.h
Naktakala: Mmmm. Good topic for a design issue. PWLD vs PWLC in this case use the same basis functions on a given cell. Maybe the solution here is to make these routines call/use something from the cell-mapping object. Let me explore this further without going out of scope here. Issue #192 created.

ChiTech/ChiMath/SpatialDiscretization/FiniteElement/PiecewiseLinear/pwlc.h
Naktakala: Issue #192 created.

ChiModules/LinearBoltzmannSolver/lbs_04c_io_flux_moments.cc
Naktakala: Done.

Naktakala:
Pinging @zachhardy @wdhawkins @dominiccaron

Naktakala:
@dominiccaron and @wdhawkins. Have your requested changes been sufficiently addressed?

ChiModules/LinearBoltzmannSolver/IterativeOperations/lbs_setsource.cc
```diff
@@ -145,22 +147,22 @@ SetSource(LBSGroupset& groupset,
                 if (not options.use_precursors)
                   infission_g += xs->chi[g] *
                                  xs->nu_sigma_f[gprime] *
-                                 phi_old_local[ir + gprime];
+                                 phi_old_local[uk_map + gprime];
 
-                //with delayed neutron precursors
+                  //with delayed neutron precursors
```
zachhardy:
```suggestion
                //with delayed neutron precursors
```

ChiModules/LinearBoltzmannSolver/IterativeOperations/lbs_setsource.cc
```diff
@@ -178,27 +180,27 @@ SetSource(LBSGroupset& groupset,
                 if (not options.use_precursors)
                   infission_g += xs->chi[g] *
                                  xs->nu_sigma_f[gprime] *
-                                 phi_old_local[ir + gprime];
+                                 phi_old_local[uk_map + gprime];
 
                   //with delayed neutron precursors
```
zachhardy:
```suggestion
                //with delayed neutron precursors
```

zachhardy (approved):
In case you want to fix some indent mismatches in comments :) If not, looks good to me.

ChiModules/LinearBoltzmannSolver/lbs_04c_io_flux_moments.cc
```diff
@@ -0,0 +1,327 @@
...
+
+#include "chi_log.h"
+extern ChiLog& chi_log;
+
+#include "chi_mpi.h"
+extern ChiMPI& chi_mpi;
+
+#include <fstream>
+#include <cstring>
+#include <assert.h>
```
dominiccaron: Same functionality, but in the interest of giving preference to C++ rather than C: `#include <cassert>`.

ChiTest/Transport3D_1Poly_qmom_part1.lua
```diff
@@ -0,0 +1,183 @@
+-- 3D Transport test with Vacuum and Incident-isotropic BC.
+-- SDM: PWLD
+-- Test: Max-value=5.27450e-01 and 3.76339e-04
```
dominiccaron: This continues to be inconsistent with the control values in the test runner.

ChiTest/Transport3D_1Poly_qmom_part2.lua
```diff
@@ -0,0 +1,182 @@
+-- 3D Transport test with Vacuum and Incident-isotropic BC.
+-- SDM: PWLD
+-- Test: Max-value=5.27450e-01 and 3.76339e-04
```
dominiccaron: As above, inconsistency with the control values in the test runner.

dominiccaron (changes requested):
Very minor changes requested, mainly regarding consistency of the test descriptions and control values in order to avoid possible confusion in the future.

ChiModules/LinearBoltzmannSolver/IterativeOperations/lbs_setsource.cc
```diff
@@ -77,15 +79,15 @@ SetSource(LBSGroupset& groupset,
...
+        size_t uk_map = full_cell_view.MapDOF(i, m, 0); //unknown map
 
-        //=================================== Loop over groupset groups
+        //============================== Loop over groupset groups
         for (size_t g = gs_i; g <= gs_f; ++g)
         {
-          //=================================== Apply material source
-          if (apply_mat_src && (ell == 0))
-            destination_q[ir + g] += src[g];
+          if ( apply_mat_src and (ell == 0) and (not options.use_src_moments))
```
wdhawkins: The paren usage here makes this a bit confusing to read.  Could we simplify it a bit?

ChiModules/LinearBoltzmannSolver/lua/lbs_io_flux_moments.cc
```diff
@@ -0,0 +1,221 @@
+#include "ChiLua/chi_lua.h"
+
+#include "../lbs_linear_boltzmann_solver.h"
```
wdhawkins: This should be fully qualified.

wdhawkins (changes requested):

ChiModules/LinearBoltzmannSolver/IterativeOperations/lbs_setsource.cc
Naktakala: The process of simplification would then be subjected to subjective evaluation. (that sounds like something from the matrix haha).
Could you tell me what you would like to see?

ChiModules/LinearBoltzmannSolver/lua/lbs_io_flux_moments.cc
Naktakala: Done

ChiTest/Transport3D_1Poly_qmom_part2.lua
Naktakala: Done

ChiTest/Transport3D_1Poly_qmom_part1.lua
Naktakala: Done

ChiModules/LinearBoltzmannSolver/lbs_04c_io_flux_moments.cc
Naktakala: Done

ChiModules/LinearBoltzmannSolver/IterativeOperations/lbs_setsource.cc
Naktakala: Done

ChiModules/LinearBoltzmannSolver/IterativeOperations/lbs_setsource.cc
Naktakala: How about this?
```c++
if (not options.use_src_moments) //using regular material src
{
  if (apply_mat_src and ell == 0)
    destination_q[uk_map + g] += src[g];
}
else if (apply_mat_src) //using ext_src_moments
  destination_q[uk_map + g] += ext_src_moments_local[uk_map + g];
```

dominiccaron (approved):

PR #185, opened 2021-07-23
Minor change to spherical harmonics

Naktakala:
## Highlights
- A minor change to the spherical harmonics implementation to make it consistent with the theory manual and PDT.
- Changed the test lua script for more concise tests (testing addition theorem and orthogonality).

## Detail - background
When looking at the wikipedia version of the real spherical harmonics there is a `sqrt{\frac{2\ell +1}{4\pi}}` term for each `Y_{\ell m}`. The corresponding harmonic expansion then does not have any `\frac{2\ell +1}{4\pi}` factor since it is embedded in the `Y_{\ell m}` and `\phi_{\ell m}` terms.
image
image
image

In transport simulations it is advantageous to have the `sqrt{\frac{2\ell +1}{4\pi}}` removed from `Y_{\ell m}` which allows the zeroth flux moment to reduce to the scalar flux (scattering goodness :-)). Therefore `Y_{\ell m}` is defined as
image
with a corresponding expansion
image
image
which gives
image

## Bottom line
Early on with ChiTech we identified that the Condon-Shortley phase factor `(-1)^m` is not required to make a proper spherical harmonic expansion and therefore ChiTech's definition of `Y_{\ell m}` is currently
image

### Pros
- Still gives the right answer.
- Still produces a valid harmonic expansion.
- Slightly less computation of the power `m` term.

### Cons
- The following identity does not hold:
image
image

## Change gives consistency with PDT, beneficial identities and diffusion theory
We are now changing the harmonic definitions to 
image

### This is the same definition that PDT used and produces the identity above

# Conclusion
- Doesn't change answers
- Improves awesomeness in theory manual


ChiTech/ChiMath/Quadratures/LegendrePoly/spherical_harmonics.cpp
```diff
@@ -10,34 +11,26 @@
...
-  {
-    if (x==0) return 1;
-    if (x==1) return 1;
-
-    int factorial = 1;
-    for (int i=2; i<=x; ++i)
-      factorial *= i;
-
-    return factorial;
-  };
```
Naktakala, line 23: This factorial has been shipped out to a utility function in `chi_math`.

Naktakala:
No takers? Really. I wrote it up so well....

ragusa:
The main change is a (-1)^m. Correct? Question:
- Have anisotropic test cases changed in value? Somehow, I was expecting a double change of (-1)^m that would cancel each other out but there is only one, so I am unsure whether results are unchanged.

Naktakala:
> The main change is a (-1)^m. Correct? Question: - Have anisotropic test cases changed in value? Somehow, I was expecting a double change of (-1)^m that would cancel each other out but there is only one, so I am unsure whether results are unchanged.

Anisotropic regression test have not changed in value. And there are indeed two of (-1)^m ... One in the Ylm used in the expansion and one in the Ylm used in the phi_lm coefficient. 
image
image

Agreed?

ragusa:
Makes sense. So they cancel each other out. I’m good.

zachhardy:
Looks good. Just needs a rebase.

zachhardy (approved):

Naktakala:
This has been waiting for a long time. @ragusa gave informal approval so I will merge this.

PR #186, opened 2021-07-26
Groupset ID and Renaming

zachhardy:
### Highlights

- Added `id` attribute to `LBSGroupset` to eliminate needs for `group_set_num` function inputs.
- Standardized the naming convention of `groupset`. The naming convention should make obvious whether `groupset` is one or two words. The options are then (changed to the first of these):
    - `LBSGroupset`, `Groupset` (directory name) `groupset`, and `groupsets`, or
    - `LBSGroupSet`, `GroupSet` (directory name), `group_set`, and `group_sets`


Naktakala (approved):

Naktakala:
This one is small enough to warrant me merging it. So I will do it in an hour or so.

PR #187, opened 2021-07-27
Added fissile cross section checks.

### LINKED ISSUE #146: Transport XS - improve consistency with Chi-prompt versus Chi

zachhardy:
### Highlights

- This PR is meant to resolve #146 
- Added various checks to ensure fission related cross sections are appropriate for the application.
- Enforce that `nu`/`chi`, `nu_prompt`/`chi_prompt`, or both are provided in all problems with fission.
- Enforce that `precursor_lambda`, `precursor_yield`, `nu_delayed`, and `chi_delayed` are provided if `num_precursors` is non-zero
-  Enforce unit `precursor_yield` and all `chi` spectra.
- If prompt and delayed quantities are provided, automatically compute `nu = nu_prompt + nu_delayed` and `chi` from `nu`- and `precursor_yield`-weighted prompt and delayed spectra.

### DIFF
--- a/ChiModules/LinearBoltzmannSolver/lbs_01b_initmaterials.cc
+++ b/ChiModules/LinearBoltzmannSolver/lbs_01b_initmaterials.cc
@@ -112,38 +112,7 @@ void LinearBoltzmann::Solver::InitMaterials(std::set<int>& material_ids)
         << options.scattering_order << "."
         << " The higher moments will therefore not be used.";
     }
-
-    //====================================== Check fission cross sections
-    if (material_xs[matid_to_xs_map[mat_id]]->is_fissile)
-    {
-      auto xs =
-          std::static_pointer_cast<chi_physics::TransportCrossSections>(
-              material_xs[matid_to_xs_map[mat_id]]);
-
-      //check what quantities have been provided
-      bool has_total = false;
-      bool has_prompt = false;
-      for (size_t g = 0; g < xs->num_groups; ++g)
-      {
-        if (xs->nu[g] > 0.0 and xs->chi[g] > 0.0)
-          has_total = true;
-        if (xs->nu_prompt[g] > 0.0 and xs->chi_prompt[g] > 0.0)
-          has_prompt = true;
-      }
-
-      //ensure correct quantities have been provided
-      if (not options.use_precursors)
-      {
-        if (not has_total and has_prompt)
-        {
-          chi_log.Log(LOG_ALLWARNING)
-              << __FUNCTION__ << ": Total nu/chi were not provided for "
-              << "a calculation without precursors. Setting nu/chi to "
-              << "the nu/chi prompt values.";
-        }
-      }
-    }
-
+    
     materials_list
       << " number of moments "
       << material_xs[matid_to_xs_map[mat_id]]->transfer_matrices.size() << "\n";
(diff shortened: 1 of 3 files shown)


ChiModules/LinearBoltzmannSolver/lbs_01b_initmaterials.cc
```diff
@@ -113,6 +113,37 @@ void LinearBoltzmann::Solver::InitMaterials(std::set<int>& material_ids)
         << " The higher moments will therefore not be used.";
     }
 
+    //====================================== Check fission cross sections
+    if (material_xs[matid_to_xs_map[mat_id]]->is_fissile)
+    {
+      auto xs =
+          std::static_pointer_cast<chi_physics::TransportCrossSections>(
+              material_xs[matid_to_xs_map[mat_id]]);
```
dominiccaron: This cast is unnecessary.

ChiModules/LinearBoltzmannSolver/lbs_01b_initmaterials.cc
```diff
@@ -113,6 +113,37 @@ void LinearBoltzmann::Solver::InitMaterials(std::set<int>& material_ids)
...
+
+      //ensure correct quantities have been provided
+      if (not options.use_precursors)
+      {
+        if (not has_total and has_prompt)
+        {
+          chi_log.Log(LOG_ALLWARNING)
+              << __FUNCTION__ << ": Total nu/chi were not provided for "
+              << "a calculation without precursors. Setting nu/chi to "
+              << "the nu/chi prompt values.";
```
dominiccaron: Please add a comment indicating where this assignment will occur (or has already occurred); otherwise, update the text of the warning message. To read a message stating explicitly that an assignment is being made and then not to know if or where the assignment occurs becomes troublesome when trying to trace the events of the code. Perhaps this check is better located where the assignment occurs?

ChiTech/ChiPhysics/PhysicsMaterial/transportxsections/prop_transportxs_01b_readfromchi.cc
```diff
@@ -5,6 +5,8 @@ extern ChiLog& chi_log;
 
 #include <string>
 
+double eps = std::numeric_limits<float>::epsilon();
```
dominiccaron: Is it intentional to use the single precision machine epsilon in the verifications being made below? If that was your intention, then I recommend to write explicitly `double eps = 1.0e-07;` or similar. The present line is valid C++ thanks to implicit conversion, but might be misinterpreted by the reader as an unintended mistake.

dominiccaron (changes requested):

ChiModules/LinearBoltzmannSolver/lbs_01b_initmaterials.cc
zachhardy: This was a redundant check. In the cross section parsing process, we check that total or prompt quantities are provided if `sigma_f` has non-zero values, i.e., `is_fissile == true`. This block was removed.

ChiModules/LinearBoltzmannSolver/lbs_01b_initmaterials.cc
zachhardy: This is true, and I'm also not sure why it was like this. Either way, this block of code has now been removed.

ChiTech/ChiPhysics/PhysicsMaterial/transportxsections/prop_transportxs_01b_readfromchi.cc
zachhardy: Modified to `double eps = 1.0e-8;`

Naktakala (approved):

dominiccaron (approved):

PR #189, opened 2021-07-27
Psi Refactor

zachhardy:
### Highlights

- This PR is based on #188 .
- This PR removes `psi_new_local` from being stored within the `LBSGroupset` class. This is done to keep the structure lightweight and to allow for the user to store multiple different versions of psi vectors in the same format without having to modify code from another solver.

### Changes to `LinearBoltmann::Solver`

- New attribute `std::vector<std::vector<double>> psi_new_local` holds angular fluxes groupset-wise. 
- Moved `ZeroAngularFluxDataStructures` from `LBSGroupset` to here for access to `psi_new_local`
- `psi_new_local` is only initialized if `options.save_angular_flux = true`.

### Changes to `LinearBoltzmann::SweepChunkPWL`

- Added `std::vector<double>& psi_new_local` attribute to hold a reference to the angular flux vector corresponding to the groupset provided to the constructor.
- Added `bool save_angular_flux` attribute defined by whether or not `psi_new_local` is empty or not.
- Added `std::vector<double>& destination_psi` to constructor

### Changes to `LinearBoltzmann::LBSGroupset`

- Removed `psi_new_local` and `psi_to_be_saved` attributes.
- Removed `ZeroAngularFluxDataStructures` routine.


ChiModules/LinearBoltzmannSolver/lbs_01e_initgroupsets.cc
```diff
@@ -17,12 +17,12 @@ void LinearBoltzmann::Solver::InitializeGroupsets()
...
     //================================================== Setup group angular flux
+    psi_new_local.emplace_back();
     if (options.save_angular_flux)
     {
       size_t num_ang_unknowns = discretization->GetNumLocalDOFs(grpset_psi_uk_man);
-      groupset.psi_to_be_saved = true;
       groupset.num_psi_unknowns_local = num_ang_unknowns;
-      groupset.psi_new_local.assign(num_ang_unknowns,0.0);
+      psi_new_local.back().assign(num_ang_unknowns, 0.0);
     }
```
dominiccaron: As the angular flux vector is no longer a member of the groupset, these operations are now out of place. Please relocate to the method `InitializeParrays`, where the allocation of the other local vectors occurs.

ChiModules/LinearBoltzmannSolver/lbs_linear_boltzmann_solver.h
```diff
@@ -186,6 +190,17 @@ class Solver : public chi_physics::Solver
...
+  //
+  void ZeroAngularFluxDataStructures(LBSGroupset& groupset)
+  {
+    if (options.save_angular_flux)
+    {
+      auto& psi = psi_new_local[groupset.id];
+      psi.assign(psi.size(), 0.0);
+    }
+    groupset.angle_agg.ZeroOutgoingDelayedPsi();
+  }
```
dominiccaron: Every segment of code that involves a call to `ZeroAngularFluxDataStructures` is followed immediately by an assignment of zero to `phi_new_local`. Can we not combine these operations into a common `ZeroFluxDataStrucures` for questions of modularity and readability? The new method might read as:
```
void ZeroFluxDataStructures(LBSGroupset& groupset)
{
  phi_new_local.assign(phi_new_local.size(), 0.0);
  if (options.save_angular_flux)
    psi_new_local[groupset.id].assign(psi_new_local[groupset.id].size(), 0.0);
  groupset.angle_agg.ZeroOutgoingDelayedPsi();
}
```
Also, there is the question as to whether or not it is necessary to zero the angular flux between iterations. In our current approach, I do not believe it to be necessary. However, if we do retain the practice of zeroing-out the downstream angular flux, then it might be useful to zero-out the downstream scalar flux in the same method.

dominiccaron (changes requested):

ChiModules/LinearBoltzmannSolver/lbs_linear_boltzmann_solver.h
zachhardy: I like this idea. A future pull request I will put up after this does more to organize the sweep preparation routines and this will be a nice addition to it.

ChiModules/LinearBoltzmannSolver/lbs_01e_initgroupsets.cc
zachhardy: Done. Additionally removed the attribute `num_psi_unknowns_local` from `LBSGroupset` for the same reason.

dominiccaron (approved):

Naktakala (approved):

PR #194, opened 2021-08-16
SweepChunk Refactor

zachhardy:
## Highlights

- Housed all sweep chunk tooling and preparation routines within `SweepChunk` (virtual) / `SweepChunkPWL` (overrides).
- Implemented `SweepChunk::ZeroDestinationPhi()`. Currently, `phi_new_local.assign(..., 0.0)` is called before sweeps. This however, is under the assumption that this is the destination phi vector within the sweep chunk. This takes away the guesswork/hard-coding.
- Added `SweepChunk::ZeroFluxDataStructures()`. This routine combines the methods that zero the destination phi, destination psi, and outgoing delayed psi into one, since they are all called back-to-back before sweeping.


ChiTech/ChiMesh/SweepUtilities/sweepchunk_base.h
```diff
@@ -40,6 +40,17 @@ class chi_mesh::sweep_management::SweepChunk
...
+  /**Sets all elements of the output vector to zero.*/
+  void ZeroDestinationPhi()
+  {
+    (*x).assign((*x).size(), 0.0);
+  }
+
+  virtual void ZeroDestinationPsi() = 0;
+  virtual void ZeroIncomingDelayedPsi() = 0;
+  virtual void ZeroOutgoingDelayedPsi() = 0;
+  virtual void ZeroFluxDataStructures() = 0;
```
dominiccaron: With the present architecture of the base class `chi_mesh::sweep_management::SweepChunk` and its derived classes, it is not ideal to declare these last four `Zero...` methods as pure virtual in this location. In particular, the methods modify class member variables that do not exist until defined in a derived class. I recommend to treat these methods in a manner analogous to get/set methods: declare and define them in the same class that defines the member variable that is modified by the method. An additional drawback of the present layout is that a derived class which does not employ the concept of a stored angular flux must override these methods nevertheless.

Two ideas come to mind:
1. move the `psi_new_local` and `groupset` (and perhaps other) member variables to the class `chi_mesh::sweep_management::SweepChunk` and define the `Zero...` methods there;
2. move the declaration of the `Zero...` methods to the class `LinearBoltzmann::SweepChunkPWL`; this may require a minor refactor of LBS to work with the derived class.

I believe that the intention of `chi_mesh::sweep_management::SweepChunk` is to be quite abstract, so likely the second option. @Naktakala, @wdhawkins?

dominiccaron (changes requested):

ChiTech/ChiMesh/SweepUtilities/sweepchunk_base.h
zachhardy: I had this same thought and discussed it with @Naktakala. This implementation was his recommendation.  

Initially, these methods were declared and implemented within `LinearBoltzmann::SweepChunkPWL`, but this led to some issues. Because we pass `chi_mesh::sweep_management::SweepScheduler` to the iterative methods and use the sweep chunk it stores of type `chi_mesh::sweep_management::SweepChunk` for sweeping, the virtual methods were needed. 

I think it adds clarity to the code to have the sweep chunk preparation routines within the sweep chunk. Unless I am misreading the suggestion for the second option, I am not sure it would work for the aforementioned reasons.

dominiccaron: I do not disagree that these methods are placed in a SweepChunk-type class, nor do I disagree that they are virtual. Rather, I disagree with the disconnect among these methods and the member variables that they are intended to modify. In addition to that point, I identify the consequences of using the pure-specifier (that means that the methods must be implemented in a derived class) in this particular scenario.

Consider the following class definition:
```
class DerivedSweepChunk : public chi_mesh::sweep_management::SweepChunk
{
  void ZeroDestinationPsi() override {}
  void ZeroIncomingDelayedPsi() override {}
  void ZeroOutgoingDelayedPsi() override {}
  void ZeroFluxDataStructures() override { ZeroDestinationPhi(); }
};
```
In other words, `DerivedSweepChunk` knows nothing about a "psi" (whether that means "angular flux" or anything else) member variable because it is not implemented it and because it is not inherited. Yet, the `Zero...` methods must be implemented in `DerivedSweepChunk` if one wishes to instantiate the class. Why require `DerivedSweepChunk` to implement methods pertaining to a member variable with which it is not concerned and which has no meaning to the class?

The previous example allows me to make another suggestion in addition to those above: it is advisable to provide a default definition of `ZeroFluxDataStructures` in the base class `chi_mesh::sweep_management::SweepChunk`. Assuming that the second approach of my original remark is adopted, the following is appropriate:
```
virtual void ZeroFluxDataStructures()
{
  ZeroDestinationPhi();
}
```

As regards the "issues" that you mention, I did warn that some additional refactor of LBS is required if the second idea is adopted. For example, that means to downcast the `chi_mesh::sweep_management::SweepChunk` object retrieved from `chi_mesh::sweep_management::SweepScheduler` to type `LinearBoltzmann::SweepChunkPWL` before calling a method that exists only in the latter. I am fairly certain that this resolves the "issues", but do please elaborate upon them if not.

zachhardy:
The most recent changes implement the routines in question above within the base class `chi_mesh::sweep_management::SweepChunk` (`SweepChunk`). This removes the need to cast to a discretization specific sweep chunk, which is also a suboptimal solution.

### Notes

- `x` renamed to `destination_phi` in `SweepChunk`
- Added `std::vector<double>* destination_psi` to `SweepChunk`
    - Added `std::vector<double>& in_destination_psi` argument to constructor.
    - Implemented analogs for all methods operating on `destination_phi` (`x`) 
- Added `AngleAggregation& angle_agg` (`chi_mesh::sweep_management::AngleAggregation`)  to `SweepChunk`
    - Added `AngleAggregation& in_angle_agg` argument to constructor.
    - `ZeroIncomingDelayedPsi` and `ZeroOutgoingDelayedPsi` are methods of the `AngleAggregation` class. Previously, these were called through the `LBSGroupset` reference passed to `LinearBoltzmann::SweepChunkPWL` (`SweepChunkPWL`). This is unnecessary. Instead, we can simply ensure that the angle aggregation from the groupset passed to `SetSweepChunk`-like routines is passed along to `SweepChunk`.
    - With this change the `SweepChunk` class holds only pointers to data vectors and a reference to an `AngleAggregation` object, which is part of the `chi_mesh::sweep_management` namespace, thus keeping its generality with expanded functionality.


What are your thoughts on this change @dominiccaron ? This is @Naktakala and I's preferred solution.

Naktakala:
> The most recent changes implement the routines in question above within the base class `chi_mesh::sweep_management::SweepChunk` (`SweepChunk`). This removes the need to cast to a discretization specific sweep chunk, which is also a suboptimal solution.
> 
> ### Notes
> * `x` renamed to `destination_phi` in `SweepChunk`
> * Added `std::vector<double>* destination_psi` to `SweepChunk`
>   
>   * Added `std::vector<double>& in_destination_psi` argument to constructor.
>   * Implemented analogs for all methods operating on `destination_phi` (`x`)
> * Added `AngleAggregation& angle_agg` (`chi_mesh::sweep_management::AngleAggregation`)  to `SweepChunk`
>   
>   * Added `AngleAggregation& in_angle_agg` argument to constructor.
>   * `ZeroIncomingDelayedPsi` and `ZeroOutgoingDelayedPsi` are methods of the `AngleAggregation` class. Previously, these were called through the `LBSGroupset` reference passed to `LinearBoltzmann::SweepChunkPWL` (`SweepChunkPWL`). This is unnecessary. Instead, we can simply ensure that the angle aggregation from the groupset passed to `SetSweepChunk`-like routines is passed along to `SweepChunk`.
>   * With this change the `SweepChunk` class holds only pointers to data vectors and a reference to an `AngleAggregation` object, which is part of the `chi_mesh::sweep_management` namespace, thus keeping its generality with expanded functionality.
> 
> What are your thoughts on this change @dominiccaron ? This is @Naktakala and I's preferred solution.

Me and Zach had a discussion on this today. The weakness here was the leanness of the base class. I think this way is more general as it acts on an `AngleAggregation` object (base ChiTech), and STL containers.

dominiccaron:
@zachhardy, @Naktakala: I do like the new structure with the angle aggregation being manipulated in the sweep-chunk by itself rather than directly off the group-set; it is more modular than the previous form.

@zachhardy: Kindly revisit commit 3f5dccc23500899651c55b38237baacd25688102; the character `x` is replaced in many more locations than intended. I will review the latest changes in detail after that matter is resolved.

A final remark, only because it is mentioned in the comments above and because it may become relevant in the future. A single discretisation-specific sweep-chunk is what we have presently in LBS, which is why it appears in the earlier discussion; however, by no means is it required to cast all the way down to a discretisation-specific sweep-chunk in order to perform LBS-specific, discretisation-independent operations. Imagine an intermediate `LinearBoltzmann::SweepChunk` class that is inserted into the inheritance hierarchy between `chi_mesh::sweep_management::SweepChunk` and `LinearBoltzmann::SweepChunkPWL`; I suppose that such a class may come to exist if/when we implement other types of spatial discretisations in the code and identify the LBS-specific, discretisation-independent common items of the sweep-chunk.

zachhardy:
> @zachhardy, @Naktakala: I do like the new structure with the angle aggregation being manipulated in the sweep-chunk by itself rather than directly off the group-set; it is more modular than the previous form.
> 
> @zachhardy: Kindly revisit commit [3f5dccc](https://github.com/chi-tech/chi-tech/commit/3f5dccc23500899651c55b38237baacd25688102); the character `x` is replaced in many more locations than intended. I will review the latest changes in detail after that matter is resolved.
> 
> A final remark, only because it is mentioned in the comments above and because it may become relevant in the future. A single discretisation-specific sweep-chunk is what we have presently in LBS, which is why it appears in the earlier discussion; however, by no means is it required to cast all the way down to a discretisation-specific sweep-chunk in order to perform LBS-specific, discretisation-independent operations. Imagine an intermediate `LinearBoltzmann::SweepChunk` class that is inserted into the inheritance hierarchy between `chi_mesh::sweep_management::SweepChunk` and `LinearBoltzmann::SweepChunkPWL`; I suppose that such a class may come to exist if/when we implement other types of spatial discretisations in the code and identify the LBS-specific, discretisation-independent common items of the sweep-chunk.

With regards to a `LinearBoltzmann::SweepChunk` class, to me, this seems like it would be over-generalization of the base class for a sweep chunk. I concur with @Naktakala that the base class was too lean and that the changes add necessary functionality without impacting the generality of the class.

My apologies for the issue with renaming `x` to `destination_phi`. The refactor button in CLion likes to be overzealous apparently. I dropped the commit and reimplemented that change. Regression tests all pass in its current state.

Naktakala (approved):

Naktakala:
@dominiccaron Are you ready to approve?

Naktakala:
@zachhardy can you rebase and make sure its mergeable then I will merge it.

zachhardy:
@Naktakala Should have posted earlier when I caught up to development, but this should be good to go. All tests passed.

PR #196, opened 2021-09-08
cmake - update minimum required version

dominiccaron:
This commit updates the minimum required version of cmake such that support is provided for the features presently employed in the build system.


Naktakala:
Just a question. Is there anyway we can support cmake version 10. It seems that TAMU-IT self-help by default installs version 10.

Naktakala:
Ping @dominiccaron

PR #197, opened 2021-09-10
Physics options

Naktakala:
## Highlights
- `chi_physics::Solver::text_name`. The base class for solvers now requires a name in its constructor. This is necessary for the coming changes to field functions. All child-classes now have to supply a name via their constructors.
- `chi_data_types::Varying`. Created a datatype called `chi_data_types::Varying` which supports basic data types `VOID`, `ARBITRARY_BYTES`, `STRING`(aka `std::string`), `BOOL`(aka `bool`), `INTEGER`(aka `int64_t`) and `FLOAT`(aka `double`). Use of this datatype is expanded below but can also be used for serialization/deserialization purposes.
- `chi_physics::BasicOption`. The `Varying` datatype is used to build a `BasicOption`-object which is essentially a pair comprising a string name and a `Varying` datapiece.
- `chi_physics::BasicOptions`. The `BasicOption` object is then used to construct a `BasicOptions`-object that is attached to the physics solver. This has the following benefits:
  - Any solver can add options to the base solver options stack with a call like `basic_options.AddOption("RESIDUAL_TOLERANCE",...)`
  - Derivative solvers now have a more structured means to define basic options.
  - The input language now has `chiSolverSetBasicOption(solver_handle, "OPTION_NAME", value)` with no knowledge needed of any child/derivative solver. **This makes it easier for mini-project developers who don't know a lot about the lua-wrappers.**
- Minor renaming and moving of files.


ChiModules/DiffusionSolver/Solver/constrdestr.cc
```diff
@@ -3,19 +3,33 @@
...
 extern ChiLog& chi_log;
 
-chi_diffusion::Solver::Solver()
-{}
+/**\defgroup LuaDiffusionBasicOptions Basic Options
+ * \ingroup LuaDiffusion
+ *
+Option name           | Type   | Default Value | Description
+----------------------|------- |---------------|------------
+discretization_method | string | 500           | Spatial discretization method.
```
zachhardy:
```suggestion
discretization_method | string | "None"           | Spatial discretization method.
```

ChiModules/KEigenvalueSolver/initialize_k_solver.cc
```diff
@@ -11,7 +11,7 @@ using namespace LinearBoltzmann;
 //###################################################################
 void KEigenvalue::Solver::InitializeKSolver()
 {
-  Initialize(); // LinearBoltzmann::Solver initialize
+  LinearBoltzmann::Solver::Initialize(); // LinearBoltzmann::Solver initialize
```
zachhardy:
```suggestion
  LinearBoltzmann::Solver::Initialize();
```

Useless comment with the change.

zachhardy (changes requested):
This is a nice addition. Just a few small comments.

- Big fan of the generalized `chiSolverInitialize` and `chiSolverExecute` routines. 
- The `diffusion_setproperty.cc` and the `DiffusionSolver`'s `lua_register` need to be cleaned up to remove references to everything now set within `BasicOptions` such as the property maps, discretization methods, residual tolerance, and max iterations.

ChiModules/DiffusionSolver/Solver/constrdestr.cc
Naktakala: Done.

ChiModules/KEigenvalueSolver/initialize_k_solver.cc
Naktakala: Done

zachhardy (approved):

PR #198, opened 2021-09-14
Updates to Cross Sections

zachhardy:
## Notes

- Changed `inv_velocity` to `velocity` and `INV_VELOCITY` to `VELOCITY` in cross section file parsing.
- Moved checks on total, prompt, delayed and absorption cross section computing terms to a new `FinalizeCrossSections` to clean up the parsing routine.


Naktakala:
Can you expand on the reason for changing from `inv_velocity` to just `velocity`? I suppose it would be easier to identify 2,200 m/s as `2.2e5` cm/s, compared to `4.54545e-6` s/cm.

zachhardy:
> Can you expand on the reason for changing from `inv_velocity` to just `velocity`? I suppose it would be easier to identify 2,200 m/s as `2.2e5` cm/s, compared to `4.54545e-6` s/cm.

This is correct. Initially, we decided to use the term `ddt_coeff` to avoid calling it a velocity. This then became `inv_velocity` when we were told it was not an issue to call it that. There are a few reasons for the change:

- Looking at benchmark problems, velocities are provided, not inverse velocities. 
- `velocity` is more intuitive to use than `inv_velocity` for the reason you gave above.
- There is no reason I see to force people to precompute the inverse velocity when the only change in the code is `* inv_velocity` -> `/ velocity`.

ragusa:
> > Can you expand on the reason for changing from `inv_velocity` to just `velocity`? I suppose it would be easier to identify 2,200 m/s as `2.2e5` cm/s, compared to `4.54545e-6` s/cm.
> 
> This is correct. Initially, we decided to use the term `ddt_coeff` to avoid calling it a velocity. This then became `inv_velocity` when we were told it was not an issue to call it that. There are a few reasons for the change:
> 
> * Looking at benchmark problems, velocities are provided, not inverse velocities.
> * `velocity` is more intuitive to use than `inv_velocity` for the reason you gave above.
> * There is no reason I see to force people to precompute the inverse velocity when the only change in the code is `* inv_velocity` -> `/ velocity`.

Spectrum codes that generate MG XS provide inverse of velocities. Not only NJOY, but other codes, too.

Why? Because a division is more expansive than a multiplication. This is still true today.

I recommend that the code computes using inv velocities. 

I recommend that the code be able to read in either velocities or inverse velocities and that the read-in data be stored as inv velocities.

zachhardy:
> > > Can you expand on the reason for changing from `inv_velocity` to just `velocity`? I suppose it would be easier to identify 2,200 m/s as `2.2e5` cm/s, compared to `4.54545e-6` s/cm.
> > 
> > 
> > This is correct. Initially, we decided to use the term `ddt_coeff` to avoid calling it a velocity. This then became `inv_velocity` when we were told it was not an issue to call it that. There are a few reasons for the change:
> > 
> > * Looking at benchmark problems, velocities are provided, not inverse velocities.
> > * `velocity` is more intuitive to use than `inv_velocity` for the reason you gave above.
> > * There is no reason I see to force people to precompute the inverse velocity when the only change in the code is `* inv_velocity` -> `/ velocity`.
> 
> Spectrum codes that generate MG XS provide inverse of velocities. Not only NJOY, but other codes, too.
> 
> Why? Because a division is more expansive than a multiplication. This is still true today.
> 
> I recommend that the code computes using inv velocities.
> 
> I recommend that the code be able to read in either velocities or inverse velocities and that the read-in data be stored as inv velocities.

This is a good point. I will make this change.

zachhardy:
The modification searches for both velocities and inverse velocities. The code should account for many different input variations:

- If velocity only is provided, `inv_velocity` is filled with the parsed values, then after parsing, the values are inverted.
- If inverse velocity is provided, done.
- If velocity and inverse velocity is provided and velocity is encountered first, `inv_velocity` is filled with the parsed velocities, then, when inverse velocities are encountered, `inv_velocitiy` is overridden.
- If velocity and inverse velocity is provided and inverse velocity is encountered first. `inv_velocity` is filled with the parsed inverse velocities and then velocities are skipped.

PR #202, opened 2021-09-16
Added volume to CellLBSView.

zachhardy:
This pull requests adds cell volumes to CellLBSView.

Delayed neutron precursors being defined at the nodes is only applicable for PWD discretizations. In PWC discretizations, precursors cannot be defined at vertices because at material interfaces there may be different (or no) precursors on each bordering cell. For generalization, the precursors should be defined at cell centers. This requires the precursor production term in the precursor equation to use the cell-averaged flux, which requires division by the cell volume. This additionally allows for easy computation of any other cell-averaged quantities that may be of interest.

NOTE: Previous PR closed due to having unrelated changes from #198 .


Naktakala (approved):
This might not be the best way to do this but for now lets go with it.

PR #203, opened 2021-09-30
Cell Centered Precursors

zachhardy:
## Notes

Currently, only the maximum number of precursors that exist on a material are stored per node. This is done to eliminate the possibility of potentially very sparse vectors in problems with many materials. With this infrastructure, nodally define precursors only make sense with PWLD and FV. With PWLC, nodally defined precursors do not make sense because interfaces between materials can have different precursor species on either side. The obvious issue here is that we have `max_precursors_per_material` slots to store up to `2 * max_precursors_per_material`. For this reason, it makes more sense to store cell averaged precursors. 

### Changes to `LinearBoltzmann::Solver`

- Added precursor initialization to `LinearBoltzmann::Solver::InitMaterials` and `LinearBoltzmann::Solver::InitPArrays`
- Added `LinearBoltzmann::Solver::ComputePrecursors` routine to compute steady state precursor concentrations.
- Call `ComputePrecursors` at the end of the `Execute` routine
- Added `num_groups`, `num_precursors`, and `max_precursors_per_material` attributes

### DIFF
--- a/ChiModules/LinearBoltzmannSolver/lbs_01b_initmaterials.cc
+++ b/ChiModules/LinearBoltzmannSolver/lbs_01b_initmaterials.cc
@@ -118,10 +118,25 @@ void LinearBoltzmann::Solver::InitMaterials(std::set<int>& material_ids)
       << material_xs[matid_to_xs_map[mat_id]]->transfer_matrices.size() << "\n";
   }//for material id
 
+  num_groups = int(groups.size());
+
   chi_log.Log(LOG_0)
     << "Materials Initialized:\n" << materials_list.str() << "\n";
 
   MPI_Barrier(MPI_COMM_WORLD);
+
+  //%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%% Initialize precursor
+  //                                                   properties
+  num_precursors = 0;
+  max_precursors_per_material = 0;
+  for (auto& xs : material_xs)
+  {
+    num_precursors += xs->num_precursors;
+    if (xs->num_precursors > max_precursors_per_material)
+      max_precursors_per_material = xs->num_precursors;
+  }
+  if (num_precursors == 0)
+    options.use_precursors = false;
 
   //%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%% Initialize Diffusion
   //                                                   properties
--- a/ChiModules/LinearBoltzmannSolver/lbs_01b_initmaterials.cc
+++ b/ChiModules/LinearBoltzmannSolver/lbs_01b_initmaterials.cc
@@ -118,7 +118,7 @@ void LinearBoltzmann::Solver::InitMaterials(std::set<int>& material_ids)
       << material_xs[matid_to_xs_map[mat_id]]->transfer_matrices.size() << "\n";
   }//for material id
 
-  num_groups = int(groups.size());
+  num_groups = groups.size();
 
   chi_log.Log(LOG_0)
     << "Materials Initialized:\n" << materials_list.str() << "\n";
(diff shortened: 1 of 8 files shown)


ChiModules/LinearBoltzmannSolver/lbs_01b_initmaterials.cc
```diff
@@ -118,11 +118,26 @@ void LinearBoltzmann::Solver::InitMaterials(std::set<int>& material_ids)
       << material_xs[matid_to_xs_map[mat_id]]->transfer_matrices.size() << "\n";
   }//for material id
 
+  num_groups = int(groups.size());
```
Naktakala: This `int(groups.size())` cast should not be necessary and its c-style. Just remove it please.

ChiModules/LinearBoltzmannSolver/lbs_compute_precursors.cc
```diff
@@ -0,0 +1,61 @@
...
+#include "chi_log.h"
+extern ChiLog& chi_log;
+
+
+//###################################################################
+/**Compute the steady state delayed neutron precursor concentrations.*/
+void LinearBoltzmann::Solver::ComputePrecursors()
+{
+  auto pwl =
+      std::dynamic_pointer_cast<SpatialDiscretization_FE>(discretization);
```
Naktakala, line 14: Since we are not casting to `SpatialDiscretization_PWL`, can we call this variable `fe` instead of `pwl`.

ChiModules/LinearBoltzmannSolver/lbs_compute_precursors.cc
```diff
@@ -0,0 +1,61 @@
...
+  const size_t J = max_precursors_per_material;
+
+  precursor_new_local.assign(precursor_new_local.size(), 0.0);
+
+  //================================================== Loop over cells
+  for (auto& cell : grid->local_cells)
```
Naktakala: Might be silly, but can you make this `const auto&`, only if it does not conflict with something.

ChiModules/LinearBoltzmannSolver/lbs_compute_precursors.cc
```diff
@@ -0,0 +1,61 @@
...
+    int xs_id = matid_to_xs_map[cell_matid];
+
+    if ((xs_id < 0) || (xs_id >= material_xs.size()))
+    {
+      chi_log.Log(LOG_ALLERROR)
+          << "Cross-section lookup error\n";
+      exit(EXIT_FAILURE);
+    }
+
+    auto xs = material_xs[xs_id];
```
Naktakala: Please change this to a const reference: `const auto& xs = *material_xs[xs_id]` and change the related arrow operators.

Naktakala (changes requested):

ChiModules/LinearBoltzmannSolver/lbs_compute_precursors.cc
zachhardy: Technically at the moment everything is `pwl` since only linear elements are supported, right (I.e. `pwld` and `pwlc` fall under `pwl`)? For anticipating future support for higher order elections, I agree with this change though.

ChiModules/LinearBoltzmannSolver/lbs_compute_precursors.cc
zachhardy: Done.

ChiModules/LinearBoltzmannSolver/lbs_compute_precursors.cc
zachhardy: Done.

ChiModules/LinearBoltzmannSolver/lbs_01b_initmaterials.cc
zachhardy: Done.

Naktakala (approved):

PR #204, opened 2021-10-01
Physics options 1 minor fixes

Naktakala:
Very minor changes. Probably will not influence anybody.


zachhardy (changes requested):
A question and a comment:

Q: How about `ComputeSigmaA_FromTransfer` (with or without the `_`) to stick with normal syntax? Another possibility is `ComputeAbsorptionFromTransfer` or `ComputeAbsorptionXSFromTransfer`.

C: In `FinalizeCrossSections`, `sigma_a` is computed as it used to be in the `MakeFromCHIxsFile` using the lambda from before. The new routine should be used there as well.

Naktakala:
> A question and a comment:
> 
> Q: How about `ComputeSigmaA_FromTransfer` (with or without the `_`) to stick with normal syntax? Another possibility is `ComputeAbsorptionFromTransfer` or `ComputeAbsorptionXSFromTransfer`.
> 
> C: In `FinalizeCrossSections`, `sigma_a` is computed as it used to be in the `MakeFromCHIxsFile` using the lambda from before. The new routine should be used there as well.

Done (`ComputeAbsorptionXSFromTransfer`) and Done.

PR #212, opened 2021-12-02
STL headers

dominiccaron:
Explicitly include utilised STL header files to permit compilation with various versions library versions. Tested for backwards compatibility on all GCC major releases down to GCC-7.


Naktakala (approved):

Naktakala:
Since this changes no code I will merge it immediately.

PR #224, opened 2022-07-08
Fixes

andrsd:


andrsd:
Hi!

I am sending 2 patches this way:
1. fixing missing `#include`s
2. fixing dependency in a `CMakeLists.txt`

I am on a MacOS in a conda environment using clang 13.0.1 - which I think is the reason why I discovered the missing `#include`s.

Not sure about the problem in `CMakeLists.txt` - I am not an expert on `cmake`. My `cmake` version is `3.19.4`.

Feel free to cherry-pick these changes - if you feel this should become part of the upstream.

Also, let me know if I can provide more insight or potentially do more testing before merging.

Thank you for consideration!

Naktakala:
I have no idea why I didn't get a notification of this. So sorry I missed this. On it now.

Naktakala:
@andrsd Ready for merging. Just one question. How did this show up for you? Reasons I am asking:
- I often fall in the trap on not including `<array>` or `<map>` or something like it.
- Do you have a special compiler or compiler flag that warns you about these?

andrsd:
I run on MacOS using conda and clang-13 (installed from conda). From what I can tell, different compiler (gcc vc clang) may have different version of std include headers, which will cause that sometimes the code would not compile on a different combination of OS and a compiler.

I have built a simple testing for ChiTech via github actions. If somebody sends a PR it would get tested on: 1) linux with clang, 2) linux with gcc and 3) macOS with clang. If you are interested I can send you a PR. This way you will have a better chance to catch these #include problems.  The longest run on the testing is about 30 minutes though (sometimes more depending if the virtual machine is overloaded). You can see how that would look like [here](https://github.com/andrsd/chi-tech/actions/runs/2677814602).

andrsd:
Correction: The linux build has only gcc because I pull in `mpich` from ubuntu package repository and that is built using gcc, rather than clang.

PR #242, opened 2022-12-02
finalizing macos instructions

ragusa:
easy install instructions for MacOS


ragusa:
somehow, this commit is also in the new diff solvers, so i am closing this PR

PR #251, opened 2022-12-18
MIP BC fix

Naktakala:


ragusa (approved):
The only thing I do not understand is the plot. The number of decade change in y and x seem to be the same on both plots. so?

Naktakala:
> The only thing I do not understand is the plot. The number of decade change in y and x seem to be the same on both plots. so?

I think this was an old PWLC plot was mislabeled.

PR #254, opened 2022-12-20
Gcc12 warnings

Naktakala:


ragusa (commented):
good to go for me. just one question. why was the chi_meshcontinuum_export file deleted ???

Naktakala:
> good to go for me. just one question. why was the chi_meshcontinuum_export file deleted ???

This method for viewing the mesh is a very old method of viewing the mesh and is very slow for any amount of cells past a few thousand (because matplotlib is slow). Nowadays it complains about string formatters for different compilers and since nobody uses it, I just ripped it out.

PR #255, opened 2022-12-21
LBS Transient solver

Naktakala:
# LBS Transient Solver

Contributed by Zachary Hardy

## Miscellaneous items
- Added a `Step` method to the solver base class for timestepping.
- `lbs::SteadySolver` old format sweepchunk removed.
- `lbs::SteadySolver::SetSource` routine removed as a virtual method and forms
  an argument for the iterative methods. 
  ClassicRichardson, GMRES and now the Krylov routines all used a `SetSource`
  routine which builds the moments-based source (among other things). The idea
  of making this virtual was such that child solver can set their custom flavor,
  which worked well up to now. The new transient solver broke this because it
  inherits from the KEigenSolver.
- Adjoint regression added.
- Corrected DSA regression tests for changes to MIP BCs.
- Moved `lbs::KEigenvalueSolver::ComputeFissionProduction` method to the based
  class, i.e., `lbs::SteadySolver`.
- `lbs::SteadySolver` source flags renamed to:
  ```c++
  enum SourceFlags : int
  {
    NO_FLAGS_SET              = 0,
    APPLY_FIXED_SOURCES       = (1 << 0),
    APPLY_WGS_SCATTER_SOURCES = (1 << 1),
    APPLY_AGS_SCATTER_SOURCES = (1 << 2),
    APPLY_WGS_FISSION_SOURCES = (1 << 3),
    APPLY_AGS_FISSION_SOURCES = (1 << 4)
  };
  ```
  The `APPLY_MATERIAL_SOURCE` was renamed `APPLY_FIXED_SOURCES` to account for 
  values do not require iterative convergence. This used to be just external 
  sources (material-, point-, or surface-sources) but now extends to include
  transient, previous timestep, precursor and angular flux values.
- `lbs::SteadySolver` now has the `ComputeFissionProduction` and 
  `ComputeFissionRate` methods. `ComputeFissionProduction` used to live under
  the KEigenSolver.
- `lbs::AdjointSolver::SetSource` is now `SetAdjointSource`.
- `lbs::SteadySolver` default sweepchunk has been replaced with a much 
  more readible source function.
- `lbs::SteadySolver` can now reinitialize materials so that it can handle XSs
  being swapped mid run.

## Transient solver
- The new transient solver is capable of simulation reactor kinetics.
- The initial time stepping method is a simple theta-scheme which supports
  backward/implicit Euler and Crank-Nicolson.
- The simulation is initialized with a KEigenValue solve to obtain steady-state-like
  initial conditions which allows one to effectively simulate step-insertions.

### Running the simulation using the `Execute` method
The execute method is very simple and can terminate on total-time or #timesteps.
In order to allow for a multitude of different methods for adaptive time-stepping
the solver allows for a callback function to be executed after each timestep 
where the user can elect to reject a time step. For example, the execute routine
is just:
```c++
void lbs::TransientSolver::Execute()
{
  chi::log.Log() << "Executing " << TextName() << ".";

  const int max_t = transient_options.max_time_steps;
  int t = 0;
  while (((max_t > 0 and t < max_t) or (max_t < 0)) and
         (time < transient_options.t_final))
  {
    Step();

    PostStepCallBackFunction();

    if (not transient_options.inhibit_advance)
    {
      AdvanceTimeValues();
      ++t;
      transient_options.inhibit_advance = false;
    }
  }

  chi::log.Log() << "Done Executing " << TextName() << ".";
}
```

One uses this callback functionality in the lua-input as follows:
```lua
swapped=false
function MyCallBack()
    FRf = chiLBSComputeFissionRate(phys1,"NEW")
    FRi = chiLBSComputeFissionRate(phys1,"OLD")
    dt = chiLBTSGetProperty(phys1, "TIMESTEP")
    time = chiLBTSGetProperty(phys1, "TIME")
    period = dt/math.log(FRf/FRi)

    if (time >= 0.2 and not swapped) then
        SwapXS(phys1, xs_weak_fuelB)
        swapped = true
    end
    chiLog(LOG_0,string.format("%s time=%10.3g dt=%10.4g period=%10.3g FR=%10.3e",
                phys1name,time,dt,period,FRf/initial_FR))
end
chiLBTSSetProperty(phys1, "CALLBACK", "MyCallBack")
chiSolverExecute(phys1)
```
In the code above here the routine `SwapXS` is a lua defined function that
swaps a materials XS.

A given timestep can be rejected by using the call
```lua
chiLBTSSetProperty(phys1, "INHIBIT_ADVANCE", true)
```
which will flip back to false when a new time step is performed.

### Running the simulation with step-by-step control
One can execute a simulation, and potentially many other solvers, by explicitly
controlling the timesteps. The code below is an example for this:
```lua
time = 0.0
time_stop = 1.0*100
k=0
swapped = false
timestep_rejected = false

tolA = 10.0
while (time < time_stop) do
    chiSolverStep(phys1)
    FRf = chiLBSComputeFissionRate(phys1,"NEW")
    FRi = chiLBSComputeFissionRate(phys1,"OLD")
    dt = chiLBTSGetProperty(phys1, "TIMESTEP")
    time = chiLBTSGetProperty(phys1, "TIME")
    period = dt/math.log(FRf/FRi)

    if (time >= 0.2 and not swapped) then
        SwapXS(phys1, xs_weak_fuelB)
        swapped = true
    end

    if (not timestep_rejected) then
        chiLBTSAdvanceTimeData(phys1)
        k = k + 1
        chiLog(LOG_0, string.format("%s %4d time=%10.3g dt=%10.4g period=%10.3g FR=%10.3e",
                phys1name,k,time,dt,period,FRf/initial_FR))
    else
        timestep_rejected = false
    end
end
```

This kind of fine-grained control allows the user to have very powerful control
over the simulation. It even allows for the coupling to other solvers and 
essentially very intricate schemes like Picard iteration.

## Tests
The simulations tested thus far as follows:
1D - 4 cells with reflective BCs (Simulating Point kinetics), initially critical, $0.21 of reactivity insert at t=0.2
image

1D - 2000 cellls, Domain size 100.0, vacuum BCs, initially critical, $0.21 of reactivity insert at t=0.2
image

The same simulations were repeated in 2D. The reflected PK-variant produced the same result. For the vacuum BC case the following geometry was used:
image

The middle material represented something like a control rod and was replaced with a more fissile material to get the exact $0.21 of reactivity. The resulting kinetic behavior was very similar to its 1D counterpart:
image

The medium- to long-time transient behavior also follows the In-hour approximation pretty well, which states that a period of ~34 seconds should asymptotically be approached in this scenario. The code follows that perfectly.


zachhardy:
Looks good to me. The two comments I have that are in the direct scope of this PR are

- I would change the `t` and `max_t` variables in `lbs::TransientSolver::Execute` to either `step` and `max_step` or `tstep` and `max_tstep`. It is a bit confusing at first glance.
- Would it be feasible to add the transients tests to `ChiTest/Z_Run_all.py`?

My probably out of scope comment is that there is inconsistency in subclass naming across the different transport solvers. The first thing is that `LBS<solver_type>Solver` is redundant and should be `LB<solver_type>Solver`. I would recommend renaming to a consistent convention. I kind of like `LBS<solver_type>` because it is shorter. With this change, I would recommend a file renaming to `lbs_<solver_name>_<operation>` for those classes affected. 

Thoughts?

Naktakala:
@zachhardy 

How does this look for you?
```c++
void lbs::TransientSolver::Execute()
{
  chi::log.Log() << "Executing " << TextName() << ".";

  const int max_num_steps = transient_options.max_time_steps;
  const double max_time = transient_options.t_final;
  int step_number = 0;
  while (((max_num_steps > 0 and step_number < max_num_steps) or
         (max_num_steps < 0)) and (time < max_time))
  {
    Step();

    PostStepCallBackFunction();

    if (not transient_options.inhibit_advance)
    {
      AdvanceTimeValues(); //new copied to prev + time+=dt
      ++step_number;
      transient_options.inhibit_advance = false;
    }
  }

  chi::log.Log() << "Done Executing " << TextName() << ".";
}
```

Naktakala:
@zachhardy Second part of your question. I will create an issue to add regression tests. I don't think the transient solver is super robust at this moment but we need the underlying architecture in the base. Also need to be tested with multigroup.

Acceptable?

zachhardy:
Sorry for the late response. Ended up having a crammed winter break because we had to drive back to DFW due to the Southwest meltdown. Everything is good with me.

PR #267, opened 2023-01-06
Adjoint Test Cleanup

zachhardy:
- There were some files left behind when running regression tests for the adjoint capabilities with the `ChiTest/Z_Run_all.py` script. This PR adds a small snippet of lua code to the end of the `Adjoint2D_#c_response.lua` scripts to remove the adjoint solutions. 

- There is additionally a small change to a CMake file that gave me some issues on my Mac. It should not result in any issues I wouldn't think. If necessary, this can be removed.


ragusa (approved):
looks like a straightforward PR. looks good to me.

Naktakala (approved):

Issue #261, opened 2022-12-23
Move to a fission matrix formulation


zachhardy:
For multi-particle transport, Chi-Tech is not currently able to handle fission in a particle agnostic way. 

Consider a neutron-gamma transport problem. Generally, this problem would be formulated as a single group structure where the first $G_n$ groups correspond to neutrons and the last $G_\gamma$ groups belong to photons. To be agnostic, the code should not have to know where one particle group structure ends and the other begins. For this example, there are fission neutron spectra for both neutron and gamma induced fission. In the current formulation, this requires the storage of the two vector spectra $\chi^{n \rightarrow n}$ and $\chi^{\gamma \rightarrow n}$. In the code, to select the correct spectra would require knowledge of which groups correspond to which particles. Instead, the aim is to shift the formulation to $\chi^{g' \rightarrow g}$ such that each incident group corresponds to an outgoing spectrum. When doing this, no knowledge of which group belongs to which particle is required. Alternatively, this can be formulated as $\sigma_f^{g' \rightarrow g} = (\chi \nu \sigma_f)^{g' \rightarrow g}$

ragusa:
I agree.


--
Jean C. Ragusa, Ph.D.
Professor, Department of Nuclear Engineering
Associate Director, Institute for Scientific Computation
Texas A&M University
Email: ***@***.******@***.***>

Websites:
http://multiphysics.engr.tamu.edu
https://nustem.engr.tamu.edu/

College Station, TX 77843-3133
Phone: (979) 862-2033
FAX: (979) 845-6443
- - - - - - - - - - - - - - - - - - - - - - - -
TEXAS A&M UNIVERSITY | FEARLESS on Every Front

On Dec 23, 2022, at 11:04 AM, Zachary Hardy ***@***.******@***.***>> wrote:

sport problem. Generally, this problem would be formulated as a single group structure where the first $G_n$ groups correspond to neutrons and the last $G_\gamma$ groups belong to photons. To be agnostic, the code should not have to know where one particle group structure ends and the other begins. For this example, there are fission neutron spectra for both neutron and gamma induced fission. In the current formulation, this requires the storage of the two vector spectra $\chi^{n \rightarrow n}$ and $\chi^{\gamma \rightarrow n}$. In the code, to select the correct spectra would require knowledge of which groups correspond to which particles. Instead, the aim is to shift the formulation to $\chi^{g' \rightarrow g}$ such that each incident group corresponds to an outg

[...]

PR #268, opened 2023-01-09
XS Cleanup + Update

zachhardy:
This PR is largely preliminary work to address #261  .

## Changes to `chi_physics::TransportCrossSections`

- Cleaned up implementation files
- Added `beta` attribute for the delayed neutron fraction
    - This allows for more flexible fission specification. Instead of requiring $\sigma_f$, $\nu^p$, and $\nu^d$, now, a users can specify $\nu\sigma_f$, $\nu$, and $\beta$, if desired. The appropriate quantities are computed when finalizing the cross-section data.
- Modified checks for fission data to be more compact and readable.

### chi_physics::TransportCrossSections

- Only allocate memory for cross-section data specified. Currently, all members are allocated regardless whether or not they are used in the simulation.
- Addition of `production_matrices` attribute to store the moment-wise group-to-group fission transfers. Several items within the discussion of #261 need to be agreed upon prior to this.

### lbs::SteadySolver

- Checks to skip areas computations involving unset cross-sections
- Setting fission sources using `production_matrices`.


ChiResources/Macros/Checks.cmake
```diff
@@ -9,7 +9,7 @@ include(CheckSymbolExists)
 CHECK_SYMBOL_EXISTS(PETSC_USE_64BIT_INDICES
                     "${PETSC_ROOT}/include/petscconf.h"
                     PETSC_USE_64BIT_INDICES)
-if (NOT PETSC_USE_64BIT_INDICES MATCHES 1)
+if (NOT ${PETSC_USE_64BIT_INDICES} MATCHES 1)
```
ragusa, line 12: didn't you already have this in the last PR? Is this from a different folder?

ChiTech/ChiPhysics/PhysicsMaterial/transportxsections/prop_transportxs_01_readfrompdt.cc
```diff
@@ -9,10 +9,13 @@
...
+  //clear any previous data
   Reset();
 
-  //======================================== Opening the file
+  //============================================================
+  // Open the file
+  //============================================================
+
   chi::log.Log()
     << "Reading PDT cross-section file \"" << file_name << "\"";
```
ragusa, line 20: do we still read PDT format XS ???

ChiTech/ChiPhysics/PhysicsMaterial/transportxsections/prop_transportxs_02_diffparams.cc
```diff
@@ -2,60 +2,85 @@
...
     }//if moment 1 available
 
-    //====================================== Determine diffcoeff
-    if (sigs_g_1 >= sigma_t[g])
+    //============================================================
+    // Compute diffusion coefficient
+    //============================================================
+
+    //display warning if sig_1 is too large
+    if (sig_1 >= sigma_t[g])
```
ragusa, line 46: we have to be very explicit in the document where the chi-tech XS format is specified. Some code provide the 2\ell+1 as part of the sigma_s,\ell. Thi comment has little bearing on the code here, but we should clearly explain what we require as input to avoid someone enter 3sig_1 instead of sig_1

ragusa, line 46: the pure forward peak scattering has all sig_s,\ell equal. If someone every try a pure forward peaked problem with ZERO absorption, then sig_total = sigs,\ell for all \ell. I am not 100% comfortable with the >= operator

ChiTech/ChiPhysics/PhysicsMaterial/transportxsections/prop_transportxs_02_diffparams.cc
```diff
@@ -2,60 +2,85 @@
...
-    //diffg[g] = 1.0/3.0/(sigma_tg[g]-sigs_g_1);
 
-    //====================================== Determine in group scattering
-    size_t num_cols = transfer_matrices[0].rowI_indices[g].size();
-    for (int j=0; j<num_cols; j++)
+    //compute the diffusion coefficient
+    diffusion_coeff[g] = 1.0 / 3.0 / (sigma_t[g] - sig_1);
+
+    //set max for when sig_t - sig_1 is near zero
+    diffusion_coeff[g] = std::fmin(1.0e12, diffusion_coeff[g]);
```
ragusa: you should merge lines 58 and 61.

diffusion_coeff[g] = 1.0 / std::fmin( 1.0e-12, 3.0 * (sigma_t[g] - sig_1 )

ChiTech/ChiPhysics/PhysicsMaterial/transportxsections/prop_transportxs_02_diffparams.cc
```diff
@@ -2,60 +2,85 @@
...
+        }
     }
 
-    //====================================== Determine removal cross-section
-    sigma_removal[g] = std::max(0.0, sigma_t[g] - sigma_s_gtog[g] - sigma_f[g]);
+    //============================================================
+    // Compute removal cross-section
+    //============================================================
+
+    sigma_removal[g] = std::max(0.0, sigma_t[g] - sigma_s_gtog[g]);
```
ragusa, line 81: I think the plain definition of sigma_removal should be used, that is, no max(0, sig_r). However, I can see your convern about round-offs.
HOWEVER, there needs to be several checks for the validity of the XS somewhere. We had plenty in PDT, including micro-reversibility and so on. Here, I am just asking that there is a test that checks that sigma_t[g] >= sigma_s_gtog[g]. Where is it?

ChiTech/ChiPhysics/PhysicsMaterial/transportxsections/prop_transportxs_05_pushlua.cc
```diff
@@ -56,22 +58,22 @@ void chi_physics::TransportCrossSections::PushLuaTable(lua_State *L)
...
+        unsigned int j = 0;
         for (auto val : row)
         {
           ++j;
-          lua_pushinteger(L,j);
-          lua_pushnumber(L,val);
-          lua_settable(L,-3);
+          lua_pushinteger(L, j);
+          lua_pushnumber(L, val);
+          lua_settable(L, -3);
```
ragusa, line 76: for my own education, what is the "-3" doing?

ChiTech/ChiPhysics/PhysicsMaterial/transportxsections/prop_transportxs_05_pushlua.cc
```diff
@@ -104,28 +106,28 @@ void chi_physics::TransportCrossSections::PushLuaTable(lua_State *L)
...
         {
           const auto& col_indices = matrix.rowI_indices[g];
           const auto& col_values  = matrix.rowI_values[g];
 
           size_t num_vals = col_values.size();
-          lua_pushinteger(L,g+1);
+          lua_pushinteger(L, g + 1);
           lua_newtable(L);
-          for (int gg=0; gg<num_vals; ++gg)
+          for (unsigned int gg = 0; gg < num_vals; ++gg)
```
ragusa, line 130: earlier, the code used gp for the other group. here, it uses gg. When other developers read code, they look for implicit clues to make a mental construction of what's going on. I think you should stick with gp and not switch to gg here, unless I missed something.

ChiTech/ChiPhysics/PhysicsMaterial/transportxsections/prop_transportxs_01c_finalize.cc
```diff
@@ -3,256 +3,272 @@
...
+  //============================================================
+  // Set the absorption cross-section, if unset
+  //============================================================
 
-  if (not transfer_matrices.empty() and (sigma_a_sum < 1.0e-28))
-  {
-    sigma_a = ComputeAbsorptionXSFromTransfer();
+  if (std::all_of(sigma_a.begin(), sigma_a.end(),
+                  [](double x) { return x == 0.0; }))
+    ComputeAbsorption();
```
ragusa, line 20: is there a check that sigma_abso is >=0 in that routine?

ChiTech/ChiPhysics/PhysicsMaterial/transportxsections/prop_transportxs_01c_finalize.cc
```diff
@@ -3,256 +3,272 @@
...
+            }
+
+            if (was_specified(nu_sigma_f))
+              for (unsigned int g = 0; g < num_groups; ++g)
+                if (nu[g] != 0.0)
+                  sigma_f[g] = nu_sigma_f[g] / nu[g];
+          } //if nu and beta specified
+          else
+            throw std::runtime_error(
+                "Invalid specification of fission data.");
```
ragusa: not precisely, number of fission neutron data. fission data is a bit vague for an error code

ChiTech/ChiPhysics/PhysicsMaterial/transportxsections/prop_transportxs_01c_finalize.cc
```diff
@@ -3,256 +3,272 @@
...
-        if (precursor_lambda[j] < eps)
+        //check and normalize prompt fission spectrum
         {
-          chi::log.Log0Error()
-            << __FUNCTION__ << ": Precursor family " << j << "'s decay "
-            << "constant must be non-zero.";
-          chi::Exit(EXIT_FAILURE);
+          //throw error if not specified
+          if (!was_specified(chi_prompt))
+            throw std::runtime_error(
```
ragusa: so chi_prompt is always required? we cannot enter just the 'steady-state' chi ?

ChiTech/ChiPhysics/PhysicsMaterial/transportxsections/prop_transportxs_01c_finalize.cc
```diff
@@ -3,256 +3,272 @@
...
-      {
-        double chi_dj_sum = 0.0;
-        for (int g = 0; g < num_groups; ++g)
-          chi_dj_sum += chi_delayed[g][j];
-        if (fabs(chi_dj_sum - 1.0) > eps and chi_dj_sum > eps)
+          //normalize the precursor yields
+          double yield_sum = std::accumulate(precursor_yield.begin(),
+                                             precursor_yield.end(), 0.0);
+          for (unsigned int j = 0; j < num_precursors; ++j)
+            precursor_yield[j] /= yield_sum;
```
ragusa: why do yields need to be normalized?
Oh wait, what are yields? Don't we have access to beta[g][j] ?

ChiTech/ChiPhysics/PhysicsMaterial/transportxsections/prop_transportxs_01c_finalize.cc
```diff
@@ -3,256 +3,272 @@
...
+        }
+
+        //compute beta-weighted total fission spectrum
```
ragusa: i like this, it is simply the definition of the steady state chi, see the note I commented this morning.

ChiTech/ChiPhysics/PhysicsMaterial/transportxsections/prop_transportxs_01c_finalize.cc
```diff
@@ -3,256 +3,272 @@
...
-      {
-        chi::log.Log0Error()
-            << __FUNCTION__ << ": Total chi must be "
-            << "provided when precursors are present.";
-        chi::Exit(EXIT_FAILURE);
-      }
+      chi::log.LogAllError()
+          << "Invalid fission data encountered.\n"
+          << "Error Message:\n" << err.what();
+      chi::Exit(EXIT_FAILURE);
```
ragusa: we have different way of exiting when an error is found. here, you use chi::Exit() but earlier, you used std::runtime_error(). Do they have the same functionality? Should we stick to one?

ChiResources/Macros/Checks.cmake
zachhardy: I'm not sure, to be honest. We can remove the change if you'd like.

ChiTech/ChiPhysics/PhysicsMaterial/transportxsections/prop_transportxs_01_readfrompdt.cc
zachhardy: Not sure, I just changed the documentation in the routine for consistency with the other changes. I've never used PDT format before.

ChiTech/ChiPhysics/PhysicsMaterial/transportxsections/prop_transportxs_02_diffparams.cc
zachhardy: This is isolated from where cross-sections are read in and is only used internally as a check for computing the transport corrected diffusion coefficient. That documentation would be best suited in the documentation of the routine that reads cross-sections. The comment is a bit overkill here as that it states the obvious.

zachhardy: This is within the code that computes the transport corrected diffusion coefficient. If we change this to `>` from `>=` we will end up with a divide by zero in line 58.

ChiTech/ChiPhysics/PhysicsMaterial/transportxsections/prop_transportxs_01c_finalize.cc
zachhardy: Not quite. If you zoom out a bit you'll see the checks are separated via
```
if (num_precursors > 0)
{
    //prompt and delayed checks
}
else
{
    //steady-state checks
}
```

See line 244 for the check for $\chi$ (steady-state).

ChiTech/ChiPhysics/PhysicsMaterial/transportxsections/prop_transportxs_01c_finalize.cc
zachhardy: If you take a look at how we implement the fission term within the `SetSource` routine, we do not use the formulation $\beta_g^j \nu_g \sigma_{f,g}$ for delayed fission, but rather $\gamma_j \nu^d_{g} \sigma_{f,g}$ where $\gamma_j$ is the fractional yield for precursor species $j$. We just use the definition $\beta_g^j = \gamma_j (\nu^d_g / \nu_g)$. The initial reason for this was because of the way NJOY spits out data. For example, NJOY spits out $\nu^p$ and $\nu^d$, but not $\beta$ and the emission spectra are defined such that $\gamma_j = \sum_g \chi^d_{g,j}$ where $\sum_j \sum_g \chi^d_{g,j} = 1$.

ChiTech/ChiPhysics/PhysicsMaterial/transportxsections/prop_transportxs_01c_finalize.cc
zachhardy: All of those `throw std::runtime_error()` are inside of a try block and the code here is the catch block. The utility of this is that I don't have to both print the error message and exit at every check. When one of the runtime errors get thrown, the code you have highlighted here gets executed and `chi::Exit` gets called. The utility of this is that I don't have to have curly braces around the million checks there are!

ChiResources/Macros/Checks.cmake
Naktakala: Inconsequential. Leave it in.

ChiTech/ChiPhysics/PhysicsMaterial/transportxsections/prop_transportxs_01_readfrompdt.cc
Naktakala: We should remove this. Created issue #269

ChiTech/ChiPhysics/PhysicsMaterial/transportxsections/prop_transportxs_02_diffparams.cc
Naktakala: There is no way a user can enter 3sig_1. If the transfer matrices (S0, S1, etc.) are supplied we check if S1 is available and compute what we need.

ChiTech/ChiPhysics/PhysicsMaterial/transportxsections/prop_transportxs_02_diffparams.cc
Naktakala: Agreed

zachhardy: Addressed in last commit.

ChiTech/ChiPhysics/PhysicsMaterial/transportxsections/prop_transportxs_01c_finalize.cc
zachhardy: Addressed in last commit.

ChiTech/ChiPhysics/PhysicsMaterial/transportxsections/prop_transportxs_01c_finalize.cc
Naktakala: Yeah. We identified this previously. Created an issue #270

ChiTech/ChiPhysics/PhysicsMaterial/transportxsections/prop_transportxs_02_diffparams.cc
Naktakala: Yeah. For pure forward this logic still works.
```c++
if (sig_1 >= sigma_t[g])
{
  sig_1 = 0.0;
  chi::log.LogAllWarning()
      << "Transport corrected diffusion coefficient failed for group "
      << g << " in call to " << __FUNCTION__ << ". "
      << "sigma_t=" << sigma_t[g] << " sigs_g_(m=1)=" << sig_1
      << ". Setting sigs_g_(m=1) to zero for this group.";
}

//compute the diffusion coefficient
//cap the value for when sig_t - sig_1 is near zero
diffusion_coeff[g] = std::fmin(1.0e12,
                               1.0 / 3.0 / (sigma_t[g] - sig_1));
```

ChiTech/ChiPhysics/PhysicsMaterial/transportxsections/prop_transportxs_01c_finalize.cc
Naktakala: Tree is nice and deep but it checks out.

ChiTech/ChiPhysics/PhysicsMaterial/transportxsections/prop_transportxs_01c_finalize.cc
zachhardy: Created issue #271 per discussion with @Naktakala

Naktakala (approved):

PR #275, opened 2023-01-11
renaming and cleaning cfem diff

ragusa:


ragusa:
I am merging this to have the true link of the files in the chi-tech main repo. Nothing much has happened: I cleaned the files, moved them to a subfolder. They were not even part of the regression tests yet...

PR #281, opened 2023-01-12
XS Allocation Fixup

zachhardy:
This PR addresses #271 , tackles an issue @Naktakala and I discussed a few days ago (initializing data that may never be used), and makes a small change to a variable name to reflect a previous change.

## Summary

- Initialize data within the read routine according to the specified sizes. 
- When combining cross-sections, we know the rules about when something is initialized. Using said rules, use checks to see if something *needs* to be combined. 
- Changed `apply_mat_src` to `apply_fixed_src` to reflect the previous name change from `APPLY_MATERIAL_SOURCES` to `APPLY_FIXED_SOURCES`.

## Notes

This should be my last PR on cross-sections before moving to the fission matrix formulation.

Issue #270, opened 2023-01-09
Transport Cross-section naming of precursor yields


Naktakala:
There are some precursor input improvements planned so we can punt problems to this issue.

zachhardy:
Any thoughts on this? I have a few ideas:

1. Use a struct to describe each precursor. It *feels* better to include the emission spectra, but this could slow the code down slightly. Currently, `chi_delayed` is accessed first by group, the by precursor index (the order of the looping in the code). Going the other way will certainly be slower if we keep the same loop structure. The alternative is to handle delayed neutrons external to the main loop of groups and to loop over precursors, destination groups, then source groups.
  ```
  struct Precursor 
  {
         double decay_constant;
         double fractional_yield;  //or yield_fraction
         std::vector<double> emission_sprectra;  //maybe
   };
  ```

2. Remove `precursor_yield` in favor of scaling the emission spectra by those values. Why remove this? The alternative I listed is exactly how NJOY specifies the delayed spectra. The definition of `chi_delayed` would then be $\sum_g \chi^d_{g,j} = \gamma_j$, and $\sum_j \gamma_j = 1$, and $\sum_j \sum_g \chi^d_{g,j} = 1$. The downfall to this strategy is that it will be slightly more expensive to compute precursor updates. This, however, is not done often.

3. The easiest solution is to simply change `precursor_yield` to `precursor_yield_fraction` and leave everything else the same.

4. We could change `precursor_lambda` to either `decay_constants` or `precursor_decay_constants`. When this was first implemented, we did not want to use `decay_constants` because there could be confusion about whether it is a decay constant for the material or precursors.

Thoughts?

ragusa:
1.  I'm fine to keep it as is for now. But tell me the lines I need to look at where you think a rework would be beneficial.
  2.  Same. Point me to the lines you recommend I look at. When I reviewed a recent PR, I remember that I was at first baffled by what the precursor yield was. Actually, it was that piece of code:
***@***.***

But I understand now I was expecting beta[g][j] and you are telling me that the yield is more naturally including in chi_delayed. So, I am not sure at this point. Maybe better to discuss live.

  1.  Likely ok with me
  2.  I think that if we ever make it to a point where a material decays, then we could discuss. I am not sure there will be a confusion as of whether one is dealing with a material or a delayed n precursor. I am in favor of decay_constant as it is more explicit. Whether you add precursor in front of it or not, I'd say no for now but I'm flexible.

[...]

PR #282, opened 2023-01-13
Precursor Struct

zachhardy:
This PR contains a change in the way we handle precursors and addressed #270. I'd like to get some thoughts before we move forward with this. We can discard, if necessary. It did not take long. I will note that this branch is based was based off of #281 before it was merged and the commit "implemented Precursor struct" is the first with the changes this PR addresses.

## Summary

Now, precursors are defined within a struct

```
struct Precursor
{
  double decay_constant = 0.0;
  double fractional_yield = 0.0;
  std::vector<double> emission_spectrum;
};
```

All of the requisite changes throughout the framework and modules have been made and the KEigenvalue regression test still passes. This allows for nice easy to read code like 

```
for (const auto& precursor : xs.precursors)
   q += precursor.emission_spectrum[g] * 
            precursor.fractional_yield * 
            xs.nu_delayed_sigma_f[gprime] * 
            phi_old_local[uk_map + gprime];
```

It is also easy to combine precursors when combining cross sections via

```
unsigned int precursor_count = 0;
for (size_t x = 0; x < xsecs.size(); ++x)
{
  for (unsigned int j = 0; j < xsecs[x]->num_precursors; ++j)
  {
    unsigned int count = precursor_count + j;
    const auto& precursor = xsecs[x]->precursors[j];

    precursors[count].decay_constant = precursor.decay_constant;
    precursors[count].fractional_yield = precursor.fractional_yield * ff_i;
    precursors[count].emission_spectrum = precursor.emission_spectrum;
  }//precursor

  precursor_count += xsecs[x]->num_precursors;
}//for x
```

## Notes

- There need to be some changes to the `PushLuaTable` routine before this is fully good to go. I have not had time to figure out the requisite changes yet.
- The `ComputeBeta` routine in `lbs::TransientSolver` seems fishy to me.


zachhardy:
Here's an update of the changes just pushed:

- Changed `PRECURSOR_LAMBDA` to `PRECURSOR_DECAY_CONSTANTS` in Chi cross section files
- Changed `PRECURSOR_YIELD` to `PRECURSOR_YIELD_FRACTIONS` in Chi cross section files
- Moved most data checks (checking values of different things) to where the data is parsed and compatibility checks to after all data is parsed. This makes the most sense to me (and reads better)
- Removed `nu`, `nu_prompt`, `nu_delayed`, and `beta` as attributes. The reason for this is because these were only used within the `TransportCrossSections` class. Instead of storing these as attributes, now, we just create some local vectors before reading the ".cxs" file and populate them if that data is within the file. Once the file is read, we use them to populate the various data members we need them for.
- Removed the `Finalize()` routine due to the large refactor of checks.

#### Notes 

Once we move to a fission matrix formulation, I plan to only carry around the fission matrices, the fission cross sections, and the production cross sections (maybe). When fission is specified in vector form (what is currently supported), we will simply parse the data using local vectors, then populate the fission matrix using an outer product.

PR #290, opened 2023-01-18
Transient Fixes

zachhardy:
This PR fixes a few previously introduced bugs discovered when running transient problems.

### Summary

- `chi` was being uniformly used in `SetTransientSource` even when prompt/delayed fission was being used. Because #281 stopped initializing member data not used in the simulation, `chi` was empty, causing a segmentation fault when using prompt/delayed fission in transient simulations when trying to access values within `chi`. This issue was fixed via
```
const double spec = options.use_precursors? xs.chi_prompt[g] : xs.chi[g];
```

- `std::for_each` was used in the `TransportCrossSections` class to try to normalize spectra data. This was not the appropriate routine to use. Now, `std::transform` is used instead.
- A call to `sys.exit()` in the regression test script `tests/Z_Run_all.py` was removed to enable the running of all regression tests.


modules/LBSTransientSolver/IterativeOperations/lbts_settransientsource.cc
```diff
@@ -115,14 +115,17 @@ void lbs::TransientSolver::
           //====================== Apply accross-groupset fission
           if (fission_avail and apply_ags_fission_src)
           {
+            const double spec = options.use_precursors?
```
ragusa, line 118: if precursors are UNIMPORTANT in my physical problem (say fast burst), should I, as a user, input data for precursors? No. So use_precursors will be F. So chi will be used. Who gives chi? I do, as a user. However, I should know that I will have to enter prompt values as if they were total (hence, I enter chi_p as chi, nu_p as nu, and so on). IS this how you envision fast burst to be run?

ragusa (approved):

modules/LBSTransientSolver/IterativeOperations/lbts_settransientsource.cc
zachhardy: @ragusa I've thought about this a bit and there are numerous ways to handle it. I will caveat this by saying that when we use a production matrix (my next PR), this check goes away. Nonetheless, this is still an important discussion when considering inputs. Currently, there are two ways to specify a fast burst where only prompt neutrons are considered.

1. Supply $\chi$ and $\nu$ with the prompt values as you suggested.
2. Specify the problem with delayed neutron precursors and use $\nu^d = 0$ for all groups.
    a. This would lead to some unnecessary computation of delayed neutron terms

The first method makes sense and the second is hacky, so it would be good to make this more robust. I think the best input options would be:

1. See 1. above.
2. Supply *only* $\nu^p$ and $\chi^p$. This will require once again re-configuring the checks done on cross sections.
3. Supply all prompt/delayed data and set `options.use_precursors = false`.

Created issue #294 for this.

Naktakala: Fine with the issue creation. Please make, if you don't object, the name `spec` here be any of the following:
- `spec_g`
- `spectrum_g`
- `chi_g`

Any of these would suffice.

zachhardy: That line goes away in the next PR I'll have that implements the production matrix. I can change it or we can just leave it. Up to you @Naktakala

Naktakala: Ok

Naktakala (approved):

Issue #294, opened 2023-01-18
Make input options for prompt only fission more robust


zachhardy:
Allow for the specification of...

1. $\nu$ and $\chi$ (already available)
2. $\nu^p$ and $\chi^p$, but not $\nu^d$, $\chi^d_j$, $\gamma_j$, and $\lambda_j$
3. All prompt and delayed data, but `options.use_precursors = false`

PR #301, opened 2023-01-24
Production matrix implementation

zachhardy:
This PR addressed #261 and #294.

### Highlights

- Added `std::vector<std::vector<double>> production_matrix` and removed `chi` and `chi_prompt` from `chi_physics::TransportCrossSections`
- Modified input checks to require either ( $\nu$, $\chi$ ) or ( $\nu^p$ , $\chi^p$ ), but not both.
- Compute production matrices from vector data via $\sigma_{f g' \rightarrow g} = \chi_g \otimes \nu\sigma_{fg'}$. This will be computed from whatever was specified from the previous bullet point.
- If production matrices are supplied...
    -  Require that the fission cross section is also available. 
    - Assert that no delayed neutron precursor data was specified (see comment below). 
    - Perform a rough check for a "reasonable" $\nu$ by requiring it be either 0 or in the range (1, 10). 0 is to allow for thresholds for photo-fission and 10 was just selected to be conservative. This check is done to ensure that the fission matrix makes general sense.
- Changed `!`, `&&`, and `||` to `not`, `and`, and `or` and cleaned up some `? :` statements.

### Notes

- Because photo-fission and neutron-induced fission have different delayed neutron data, there is no way to use vector delayed neutron data without checks for particle types. For this reason, when production matrices are specified (the only was to specify multi-particle data), delayed neutron precursors are not allowed. No plans to implement this feature in the near future.
- When a production matrix is specified, $\nu\sigma_f$ is computed from it. Because no delayed neutrons are allowed, this will be either $\nu^{total}\sigma_f$ or $\nu^{prompt}\sigma_f$. In either case, the result is stored in `nu_sigma_f`. This will remain this way until delayed neutron data is implemented for production matrix specification. At present, neither `nu_sigma_f` nor `nu_prompt_sigma_f` are used in the LBS solvers.


framework/ChiPhysics/PhysicsMaterial/transportxsections/prop_transportxs_06_export.cc
```diff
@@ -145,20 +145,33 @@ void chi_physics::TransportCrossSections::
...
+                << g << " "
+                << col_values[k] << "\n";
       }//for g
 
       ofile << "\n";
     }//for ell
     ofile << "TRANSFER_MOMENTS_END\n";
   }//if has transfer matrices
 
+  if (!production_matrix.empty())
```
ragusa, line 160: ! --> not

modules/LBSTransientSolver/IterativeOperations/lbts_settransientsource.cc
```diff
@@ -70,157 +70,125 @@ void lbs::TransientSolver::
...
-          if (not options.use_src_moments) //using regular material src
-          {
-            if (apply_fixed_src and ell == 0)
-              destination_q[uk_map + g] += src[g];
-          }
+          double rhs = 0.0;
+
+          //============================== Apply fixed sources
+          if (!options.use_src_moments) //using regular material src
+            rhs += apply_fixed_src && ell == 0? src[g] : 0.0;
```
ragusa: I would add a space before the ? for readability

modules/LBSTransientSolver/IterativeOperations/lbts_settransientsource.cc
```diff
@@ -70,157 +70,125 @@ void lbs::TransientSolver::
...
+          //============================== Apply scattering sources
           const bool moment_avail = (ell < S.size());
 
-          //====================== Apply across-groupset scattering
-          if (moment_avail and apply_ags_scatter_src)
-            for (const auto& [row_g, gprime, sigma_sm] : S[ell].Row(g))
-              if ((gprime < gs_i) or (gprime > gs_f))
-                inscatter_g += sigma_sm * phi_old_local[uk_map + gprime];
+          //==================== Across groupset
+          if (moment_avail && apply_ags_scatter_src)
```
ragusa: && --> and

modules/LBSTransientSolver/IterativeOperations/lbts_settransientsource.cc
```diff
@@ -70,157 +70,125 @@ void lbs::TransientSolver::
...
+              if (gp < gs_i || gp > gs_f)
+                rhs += sigma_sm * phi_old_local[uk_map + gp];
 
-          //====================== Apply within-groupset scattering
-          if (moment_avail and apply_wgs_scatter_src)
-            for (const auto& [row_g, gprime, sigma_sm] : S[ell].Row(g))
-              if ((gprime >= gs_i) and (gprime <= gs_f))
-                inscatter_g += sigma_sm * phi_old_local[uk_map + gprime];
+          //==================== Within groupset
+          if (moment_avail && apply_wgs_scatter_src)
```
ragusa: and

modules/LBSTransientSolver/IterativeOperations/lbts_settransientsource.cc
```diff
@@ -70,157 +70,125 @@ void lbs::TransientSolver::
...
+            for (const auto& [_, gp, sigma_sm] : S[ell].Row(g))
+              if (gp >= gs_i && gp <= gs_f)
```
ragusa: and

modules/LBSTransientSolver/IterativeOperations/lbts_settransientsource.cc
```diff
@@ -70,157 +70,125 @@ void lbs::TransientSolver::
...
+                rhs += sigma_sm * phi_old_local[uk_map + gp];
 
-          destination_q[uk_map + g] += inscatter_g;
+          //============================== Apply fission sources
+          const bool fission_avail = xs.is_fissionable && ell == 0;
```
ragusa: and

modules/LBSTransientSolver/IterativeOperations/lbts_settransientsource.cc
```diff
@@ -70,157 +70,125 @@ void lbs::TransientSolver::
...
 
-          double infission_g = 0.0;
-          const bool fission_avail = (xs.is_fissionable and ell == 0);
-
-          //====================== Apply accross-groupset fission
-          if (fission_avail and apply_ags_fission_src)
+          //==================== Across groupset
+          if (fission_avail && apply_ags_fission_src)
```
ragusa: and

modules/LBSTransientSolver/IterativeOperations/lbts_settransientsource.cc
```diff
@@ -70,157 +70,125 @@ void lbs::TransientSolver::
...
           {
-            const double spec = options.use_precursors?
-                xs.chi_prompt[g] : xs.chi[g];
-
-            //=============== Loop over groups
-            for (size_t gprime = first_grp; gprime <= last_grp; ++gprime)
-              if ((gprime < gs_i) or (gprime > gs_f))
+            const auto& prod = xs.production_matrix[g];
+            for (size_t gp = first_grp; gp <= last_grp; ++gp)
+              if (gp < gs_i || gp > gs_f)
```
ragusa: or

modules/LBSTransientSolver/IterativeOperations/lbts_settransientsource.cc
```diff
@@ -70,157 +70,125 @@ void lbs::TransientSolver::
...
 
-          //====================== Apply across-groupset scattering
-          if (moment_avail and apply_ags_scatter_src)
-            for (const auto& [row_g, gprime, sigma_sm] : S[ell].Row(g))
-              if ((gprime < gs_i) or (gprime > gs_f))
-                inscatter_g += sigma_sm * phi_old_local[uk_map + gprime];
+          //==================== Across groupset
+          if (moment_avail && apply_ags_scatter_src)
+            for (const auto& [_, gp, sigma_sm] : S[ell].Row(g))
+              if (gp < gs_i || gp > gs_f)
```
ragusa: or

modules/LBSAdjointSolver/IterativeOperations/lbsadj_setsource.cc
```diff
@@ -44,106 +44,77 @@ void lbs_adjoint::AdjointSolver::
...
+                           precursor.fractional_yield *
+                           xs.nu_delayed_sigma_f[gp] *
+                           phi_old_local[uk_map + gp];
               }
-            }//for gprime
-          }//if zeroth moment
+          }
 
-          //====================== Apply within-groupset fission
+          //==================== Within groupset
```
ragusa, line 99: and

modules/LBSTransientSolver/IterativeOperations/lbts_settransientsource.cc
```diff
@@ -70,157 +70,125 @@ void lbs::TransientSolver::
...
-                                   cell_volume;
-                }//for gprime
-              }//for precursors
-            }//if use precursors
-          }//if fission_avail and apply_ags_fission_src
-
-          //====================== Apply within-groupset fission
-          if (fission_avail and apply_wgs_fission_src)
+          //==================== Within groupset
+          if (fission_avail && apply_wgs_fission_src)
```
ragusa: and

modules/LBSTransientSolver/IterativeOperations/lbts_settransientsource.cc
```diff
@@ -70,157 +70,125 @@ void lbs::TransientSolver::
...
+                         precursor.fractional_yield *
+                         xs.nu_delayed_sigma_f[gp] *
+                         phi_old_local[uk_map + gp] /
+                         cell_volume;
+                }
+            }
+          }
+
+          //============================== Apply previous precursors
+          if (fission_avail && apply_fixed_src && options.use_precursors)
```
ragusa: and

modules/LBSTransientSolver/IterativeOperations/lbts_settransientsource.cc
```diff
@@ -70,157 +70,125 @@ void lbs::TransientSolver::
...
 
         }//for g
       }//for m
     }//for dof i
   }//for cell
 
   //================================================== Apply point sources
-  if ((not options.use_src_moments) and apply_fixed_src)
-  {
+  if (!options.use_src_moments && apply_fixed_src)
```
ragusa: and

modules/LBSTransientSolver/lbts_transient_solver.h
```diff
@@ -17,13 +17,26 @@ class TransientSolver : public KEigenvalueSolver
 {
 public:
   chi_math::SteppingMethod method = chi_math::SteppingMethod::CRANK_NICHOLSON;
+
+  /// Options for initial condition normalization
+  enum class NormalizationMethod
+  {
+    TOTAL_POWER = 0,    ///< Total reactor power
+    POWER_DENSITY = 1,  ///< Power density
+    NONE = 2            ///< No normalization
```
ragusa, line 26: shouldn't NONE be the 0, seems more logical

modules/LinearBoltzmannSolver/IterativeOperations/lbs_compute_fission_production.cc
```diff
@@ -30,9 +30,8 @@ double SteadySolver::ComputeFissionProduction(const std::vector<double>& phi)
       const double IntV_ShapeI = cell_matrices.Vi_vectors[i];
 
       //=============================== Loop over groups
-      //TODO: This should be nu
       for (size_t g = first_grp; g <= last_grp; ++g)
-        local_production += xs.nu_prompt_sigma_f[g] *
+        local_production += xs.nu_sigma_f[g] *
```
ragusa: i suppose this routine is in LB steady state. so nu.sigma is good. I see that before it had nu_prompt.sigma. I think it is more readable like this now

modules/LinearBoltzmannSolver/IterativeOperations/lbs_setsource.cc
```diff
@@ -55,16 +55,14 @@ void lbs::SteadySolver::
 
     //==================== Obtain src
     double* src = default_zero_src.data();
-    if (P0_src and apply_fixed_src)
+    if (P0_src && apply_fixed_src)
```
ragusa: I am not sure why the and (old) was changed to && (new). I thought we decided to go with more english-like symbols

modules/LinearBoltzmannSolver/IterativeOperations/lbs_setsource.cc
```diff
@@ -74,103 +72,77 @@ void lbs::SteadySolver::
...
-          }
-          else if (apply_fixed_src)  //using ext_src_moments
-            destination_q[uk_map + g] += ext_src_moments_local[uk_map + g];
+          double rhs = 0.0;
 
-          double inscatter_g = 0.0;
-          const bool moment_avail = (ell < S.size());
+          //============================== Apply fixed sources
+          if (!options.use_src_moments) //using regular material src
+            rhs += apply_fixed_src && ell == 0? src[g] : 0.0;
```
ragusa: and

modules/LinearBoltzmannSolver/IterativeOperations/lbs_setsource.cc
```diff
@@ -74,103 +72,77 @@ void lbs::SteadySolver::
...
+          //============================== Apply scattering sources
+          const bool moment_avail = ell < S.size();
 
-          //====================== Apply within-groupset scattering
-          if (moment_avail and apply_wgs_scatter_src)
-            for (const auto& [row_g, gprime, sigma_sm] : S[ell].Row(g))
-              if ((gprime >= gs_i) and (gprime <= gs_f))
-                inscatter_g += sigma_sm * phi_old_local[uk_map + gprime];
+          //==================== Across groupset
+          if (moment_avail && apply_ags_scatter_src)
```
ragusa: and

modules/LinearBoltzmannSolver/IterativeOperations/lbs_setsource.cc
```diff
@@ -74,103 +72,77 @@ void lbs::SteadySolver::
...
+          //==================== Within groupset
+          if (moment_avail && apply_wgs_scatter_src)
+            for (const auto& [_, gp, sigma_sm] : S[ell].Row(g))
+              if (gp >= gs_i && gp <= gs_f)
+                rhs += sigma_sm * phi_old_local[uk_map + gp];
 
-          double infission_g = 0.0;
-          const bool fission_avail = (xs.is_fissionable and ell == 0);
+          //============================== Apply fission sources
+          const bool fission_avail = xs.is_fissionable && ell == 0;
```
ragusa: and

modules/LinearBoltzmannSolver/IterativeOperations/lbs_setsource.cc
```diff
@@ -74,103 +72,77 @@ void lbs::SteadySolver::
...
 
-          //====================== Apply accross-groupset fission
-          if (fission_avail and apply_ags_fission_src)
+          //==================== Across groupset
+          if (fission_avail && apply_ags_fission_src)
```
ragusa: and

modules/LinearBoltzmannSolver/IterativeOperations/lbs_setsource.cc
```diff
@@ -74,103 +72,77 @@ void lbs::SteadySolver::
...
+                           phi_old_local[uk_map + gp];
               }
-            }//for gprime
-          }//if zeroth moment
+          }
 
-          //====================== Apply within-groupset fission
-          if (fission_avail and apply_wgs_fission_src)
+          //==================== Within groupset
+          if (fission_avail && apply_wgs_fission_src)
```
ragusa: and

modules/LinearBoltzmannSolver/IterativeOperations/lbs_setsource.cc
```diff
@@ -74,103 +72,77 @@ void lbs::SteadySolver::
...
 
         }//for g
       }//for m
     }//for dof i
   }//for cell
 
   //================================================== Apply point sources
-  if ((not options.use_src_moments) and apply_fixed_src)
-  {
+  if (!options.use_src_moments && apply_fixed_src)
```
ragusa: and

ragusa (commented):
first pass through. a lot of times, a "&&" was changed to "and". I thoughts we wanted to go the other way

modules/LBSTransientSolver/IterativeOperations/lbts_settransientsource.cc
zachhardy: I've always used `cond? true : false` and sometimes `(cond)? true : false`. We should see how these are written in the rest of the code.

ragusa:
Here is a situation where ( ) would improve readability, imo.

modules/LinearBoltzmannSolver/IterativeOperations/lbs_compute_fission_production.cc
```diff
@@ -30,9 +30,8 @@ double SteadySolver::ComputeFissionProduction(const std::vector<double>& phi)
...
                             phi[uk_map + g] *
                             IntV_ShapeI;
     }//for node
```
Naktakala, line 46: Should this not be something like this:
```c++
      if (not options.use_precursors)
      {
        for (size_t g = first_grp; g <= last_grp; ++g)
          local_production += xs.nu_sigma_f[g] *
                              phi[uk_map + g] *
                              IntV_ShapeI;
      }
      else
      {
        for (size_t g = first_grp; g <= last_grp; ++g)
          local_production += xs.nu_prompt_sigma_f[g] *
                              phi[uk_map + g] *
                              IntV_ShapeI;
        for (const auto& precursor : xs.precursors)
          for (size_t g = first_grp; g <= last_grp; ++g)
            local_production += precursor.fractional_yield *
                                xs.nu_delayed_sigma_f[g] *
                                phi[uk_map + g] *
                                IntV_ShapeI;
      }
```

zachhardy: Hmmm. I need to think about this. Normally, these are equivalent because $\nu = \nu^p + \nu^d$ and $\sum_j \gamma_j = 1$. However, this may need to be changed.

zachhardy: Changed to 

```c++
for (size_t g = first_grp; g <= last_grp; ++g)
      {
        const auto& prod_g = xs.production_matrix[g];
        for (size_t gp = 0; gp <= last_grp; ++gp)
          local_production += prod_g[gp] *
                              phi[uk_map + gp] *
                              IntV_ShapeI;

        if (options.use_precursors)
          for (unsigned int j = 0; j < xs.num_precursors; ++j)
            local_production += xs.nu_delayed_sigma_f[g] *
                                phi[uk_map + g] *
                                IntV_ShapeI;
      }
```

Using the production matrix we don't need addition checks because it is always either total or prompt, depending on what was specified. If total was specified, `options.use_precursors` will always be false.

ragusa (approved):

modules/LinearBoltzmannSolver/IterativeOperations/lbs_compute_fission_production.cc
Naktakala: BUG! @zachhardy The two routines in this file need the following check:
```c++
const bool fission_avail = xs.is_fissionable
```
and use just like we do in the `SetSource` routines, otherwise they crash if a material is encountered without fission. I found this bug by running `tests/BigTests/c5g7/X_Test1.lua`.

Naktakala: I was able to fix this from my end.

Naktakala (approved):

PR #302, opened 2023-01-24
Extrusion upgrades

Naktakala:
## Highlights
- Added two new options for setting material- and boundary-ids:
  - `chiVolumeMesherSetProperty(MATID_FROM_LUA_FUNCTION,...`
  - `chiVolumeMesherSetProperty(BNDRYID_FROM_LUA_FUNCTION,...`
- The material-id function gets called for each cell and is passed the cell's centroid and current mat-id
- The boundary-id function gets called for each boundary face and is passed the face's centroid, face's normal and
current boundary id

- Also modified the extruded to always copy (by default) the 2D template's mat-ids.
- Various small cosmetic changes.

### DIFF
--- a/framework/ChiMesh/SurfaceMesh/lua/surfacemesh_load.cc
+++ b/framework/ChiMesh/SurfaceMesh/lua/surfacemesh_load.cc
@@ -1,24 +1,47 @@
 #include "ChiLua/chi_lua.h"
 
-#include <iostream>
 #include "ChiMesh/SurfaceMesh/chi_surfacemesh.h"
-#include "../../MeshHandler/chi_meshhandler.h"
+#include "ChiMesh/MeshHandler/chi_meshhandler.h"
 
 #include "chi_runtime.h"
-
 #include "chi_log.h"
-;
-
 
 //############################################################################# Create Window
 /** Loads mesh data from a wavefront object.
  *
 \param SurfaceHandle int Handle to the surface on which the operation is to be performed.
-\param FileName char* Path to the file to be imported.
+\param FileName string Path to the file to be imported.
 \param polyflag bool (Optional)Flag indicating whether triangles
-                               are to be read as polygons. [Default: true].
+                               are to be read as polygons. [Default: true (read
+                               as polygons)].
 \param transform table3 (Optional) Translation vector to move all the vertices.
                                    [Default: none].
+
+### Note:
+If the intent of a surface mesh is to serve as a 3D logical volume then
+the `polyFlag` parameter should be set to false.
+
+### Example
+Example usage:
+\code
+-- Basic example
+surfmesh1 = chiSurfaceMeshCreate()
+chiSurfaceMeshImportFromOBJFile(surfmesh1, "MeshFile1.obj")
+
+-- Surface mesh used as Logical volume
+lv_surfmesh1 = chiSurfaceMeshCreate()
+chiSurfaceMeshImportFromOBJFile(lv_surfmesh1, "MeshFile3D.obj", false)
+
+lv1 = chiLogicalVolumeCreate(SURFACE, lv_surfmesh1)
+
+-- Surface mesh with transform
+dx = 1.5
+dy = -2.5
+lv_surfmesh2 = chiSurfaceMeshCreate()
+chiSurfaceMeshImportFromOBJFile(lv_surfmesh2, "MeshFile3D.obj", false, {dx,dy,0.0})
+
+lv2 = chiLogicalVolumeCreate(SURFACE, lv_surfmesh2)
+\endcode
 
 \return success bool Return true if file was successfully loaded and false
  otherwise.
--- a/framework/ChiMesh/VolumeMesher/chi_volumemesher.h
+++ b/framework/ChiMesh/VolumeMesher/chi_volumemesher.h
@@ -11,25 +11,27 @@
 
 namespace chi_mesh
 {
-  enum VolumeMesherType
+  enum class VolumeMesherType
   {
     EXTRUDER      = 4,
     UNPARTITIONED = 6
   };
   enum VolumeMesherProperty
   {
-    FORCE_POLYGONS      = 1,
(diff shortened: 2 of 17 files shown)


framework/ChiMesh/SurfaceMesh/lua/surfacemesh_load.cc
```diff
@@ -1,25 +1,48 @@
...
 \param polyflag bool (Optional)Flag indicating whether triangles
-                               are to be read as polygons. [Default: true].
+                               are to be read as polygons. [Default: true (read
+                               as polygons)].
 \param transform table3 (Optional) Translation vector to move all the vertices.
                                    [Default: none].
 
+### Note:
+If the intent of a surface mesh is to serve as a 3D logical volume then
+the `polyFlag` parameter should be set to false.
```
ragusa, line 22: thanks for adding this useful comment (and many others). I was wondering about that bool

framework/ChiMesh/VolumeMesher/chi_volumemesher.h
```diff
@@ -11,25 +11,27 @@
...
+    PARTITION_Z               = 3,
+    PARTITION_Y               = 4,
+    PARTITION_X               = 5,
+    CUTS_Z                    = 6,
+    CUTS_Y                    = 7,
+    CUTS_X                    = 8,
+    PARTITION_TYPE            = 9,
+    EXTRUSION_LAYER           = 10,
+    MATID_FROMLOGICAL         = 11,
+    BNDRYID_FROMLOGICAL       = 12,
```
ragusa, line 32: I know I'm a pain. Wouldn't FROM_LOGICAL read better?

framework/ChiMesh/VolumeMesher/chi_volumemesher_01c_set_IDs.cc
```diff
@@ -109,4 +110,223 @@ SetMatIDToAll(int mat_id)
...
+    lua_pushnumber(L, xyz.y);
+    lua_pushnumber(L, xyz.z);
+    lua_pushinteger(L, cell.material_id);
+
+    //============= Call lua function
+    //4 arguments, 1 result (double), 0=original error object
+    int lua_return;
+    if (lua_pcall(L,4,1,0) == 0)
+    {
+      LuaCheckNumberValue(fname, L, -1);
```
ragusa, line 163: this fname is what I was missing in my lua calls in CFEM/DFEM in order not to hard-code the funct name. Now I know how to improve those modules. thanks

ragusa (approved):
This is really cool. Thanks for doing this so quickly (and neatly!)

framework/ChiMesh/VolumeMesher/chi_volumemesher.h
zachhardy: Is there a reason `VolumeMesherType` is an `enum class` but `VolumeMesherProperty` is only an `enum`?

Also I concur, FROM_LOGICAL would be better just as we are using FROM_LUA

framework/ChiMesh/VolumeMesher/chi_volumemesher_01_utils.cc
```diff
@@ -1,6 +1,13 @@
 #include "chi_volumemesher.h"
 #include "ChiMesh/MeshContinuum/chi_meshcontinuum.h"
 
+//###################################################################
+chi_mesh::VolumeMesher::VolumeMesher(VolumeMesherType type) :
+ m_type(type)
+{
+
+}
+
```
zachhardy: A lot of space to take up for some empty braces. Thoughts on putting `{ }` or `{}` on line 7?

zachhardy (approved):

Naktakala:
General comment for @zachhardy and @ragusa.

`VolumeMesherType` used to be just an `enum`. I have converted it to `enum class` and made it a member of `chi_mesh::VolumeMesher` class. This member is initialized in the constructor. A member function is then used to obtain the type, i.e. `mesher.Type()`. This removed RTTI calls in the `SetProperty` routine.

Convention on the use of `enum` and `enum class`:
- Caution should be exercised with `enum`s since they can be equated, e.g., in the old code `VolumeMesherProperty::PARTITION_Y == VolumeMesherType::EXTRUDER` would return true, whereas now such a statement will not compile. In fact, within the `chi_mesh` namespace the statement `PARTITION_Y == EXTRUDER` would've worked.
- Straight `enums` can also be equated to `int`, i.e., `VolumeMesherProperty::PARTITION_Y == 4` will return true.
- `enum` should be used as an efficient proxy for strings, e.g., `thing == BIG` is more efficient than `thing == "small"`, or when using integers as options, e.g. `property == PARTITION_Y` is more readable than `property == 4`.
- Whenever a type can be subjected to implicit conversion, `enum class` should be preferred. With the new code, `mesher.Type() == 4` or `mesher.Type() == EXTRUDER`, will not work. It has to be `mesher.Type() == VolumeMesherType::EXTRUDER`.
- In general it is also good practice to use `enum class` when the type is used as a class/struct member.

framework/ChiMesh/VolumeMesher/chi_volumemesher.h
Naktakala: See general comment on first one.

So this PR added a new feature and 17 files were changed. If we make the change to `FROM_LOGICAL` then ALL the regression tests need to have this statement changed. We would effectively deprecate a piece of the input language which sounds a lot like "out-of-scope" for this PR. 

I suggest that we do 2 things, i) create an issue titled "Renaming Input Language items" where we list all the input language related items that need to change. ii) When i) is done, we make a mass renaming PR.

framework/ChiMesh/VolumeMesher/chi_volumemesher_01_utils.cc
Naktakala: Done

zachhardy (approved):

PR #303, opened 2023-01-24
Cleanup and improve regression test script

zachhardy:
### Highlights

- Use `argparse` library to handle command line inputs. While the `"tests_to_run=[*range(first, last)]`" or `"tests_to_run=[test1, test2, ..., testN]"` format works, the `argparse` library is nice because it generates a `--help` option to display documentation. Now, there are two optional inputs: 1) `--test-list test1 test2 ... testN` and 2)`--test-range test_start test_end`. Each of these do exactly what they suggest and an error is thrown if both are specified.
- General comment cleanup and code cleanup.

### DIFF
--- a/tests/Z_Run_all.py
+++ b/tests/Z_Run_all.py
@@ -3,9 +3,35 @@ import os
 import sys
 import shutil
 import textwrap
+import argparse
 
-# To run a range of tests pass a text string via the command line.
-# i.e., "tests_to_run=[*range(14,16)]"
+parser = argparse.ArgumentParser(
+    description="The regression tests.",
+    formatter_class=argparse.RawTextHelpFormatter
+)
+
+parser.add_argument(
+    '--test-list',
+    nargs='*',
+    type=int,
+    required=False
+)
+
+parser.add_argument(
+    '--test-range',
+    nargs=2,
+    default=None,
+    type=int,
+    required=False
+)
+
+argv = parser.parse_args()
+
+# To run a particular set of tests, pass "--test-list" followed by space
+# separated integers corresponding to the test numbers to run.
+
+# To run a particular range of tests, pass "--test-range" followed by
+# the first test desired and the last test desired.
 
 # This python script executes the regression test suite.
 # In order to add your own test, copy one of the test blocks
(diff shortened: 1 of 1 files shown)


Naktakala:
Nooooooo! All the designation comments were removed (i.e., #13 #14 #15, etc)? Why? They bothered you? How do I know how to run a particular test or when I put in a new test, what its number would be?

zachhardy:
Well, because all of the numbers were wrong to start with. If we add a test, every single number has to be changed. I put the test number range underneath each of the test type headers instead. That at least makes it more manageable when tests are added. I can add the numbers back if you want. I'd like to eventually be able to specify a list of file paths, but that's for another day

ragusa:
Sorry guys. Long day. When I added test, I didn’t want to mess up the existing numbering so I didn’t change the comments.

Jan and I talked a while back that we needed to associate a unique number with each test in a way that is not disruptive.
What I had proposed is the the arguments for run_test be all collected in 3 lists. Each row of these 3 lists refer to one test. The list would be for (1) file to run, (2) comment to print, (3) golden value.

We didn’t like it. Too hacky. Now, I think I could do it better with a struct (a fake class). Do you want me to give it a try. In the end I want to retain the ability to down select which tests to run based on their ID number and I want new tests to be simple to add.

Let me know when I case proceed.

ragusa (commented):
I looked at it now. The numbers are back in. So, are we good? I like this. @Naktakala what do you think?

zachhardy:
I think at some level it would be nice to be able to run a set of tests based on keywords rather than numbers. An easy start would be to specify a directory or directories via
```
python test/Z_Run_all.py --test-dirs test/Transport_Steady test/Transport_Keigen
```
or something like that.

ragusa:
I like the per folder idea so let me add to it:
--test-dirs : either "all" or a list of subdirs
and how about we add a range per subdir and the numbering is done PER subdir.

Examples:
--test-dirs all                                                          --> all tests
--test-dirs test/Transport_Steady                           --> all tests in that folder
--test-dirs test/Transport_Steady --range [1,2,5]   --> only tests 1 2 and 5 of that folder
--test-dirs test/Transport_Steady --range [1,2,5]   --test-dirs test/Transport_Keigen --range [3]   --> only tests 1 2 or 5 of the first folder and test 3 of the second folder

Thoughts? @zachhardy @Naktakala

zachhardy:
I like that idea, should the default behavior be all tests _**and**_ have the `--test-dirs-all` option?

ragusa:
I guess we can have:
no arguments provided --> run all tests by default
and then, if arguments are given, we do as I proposed in the previous post

Naktakala:
> I guess we can have: no arguments provided --> run all tests by default and then, if arguments are given, we do as I proposed in the previous post

This sounds good to me. Don't feel pressured to do that NOW. You can do it in a new PR.

zachhardy:
Yea I just started looking at it and that is a bigger job. I'll work on it while I have time when I'm waiting on other simulations to finish up. It'll be a new PR and I will refresh the entire regression suite script.

PR #305, opened 2023-01-30
Deprecate old gmres

Naktakala:
This has been replaced, a while ago, with the more versatile `KRYLOV` methods that support `RICHARDSON`, `GMRES` and `BICGSTAB`.

The change does not really affect anything other than all the regression tests that used the old language.


zachhardy (approved):
Good to go. I'd just like to know the reason for changing some of the golden values on a few regression tests.

Naktakala:
> Good to go. I'd just like to know the reason for changing some of the golden values on a few regression tests.

Ah good question. The main reason for this, as far as I could tell, was that the old GMRES applied slightly more operations to the floating point values, allowing a little more round-off to creep in (judging by the difference only creeping in when the residual is fairly small). The new general KRYLOV methods now implement WGDSA and TGDSA as a single composite preconditioner. This allows PETSc to hand it a delta-phi vector directly, as compared to the old GMRES which had to convert a krylov vector to phi_old and phi_new, subtract them, then pass delta_phi to compute b, solve to get a new delta_phi, add to phi_new, then assemble another petsc vector. So overall less additions and subtractions.

PR #309, opened 2023-02-22
Common gs linsolvers

Naktakala:
# Large-scale overhaul of the linear solvers in LBS

## Highlights
- There is now a `LBSSolver` base class allowing many solvers to share the basic groupset structures.
- Added Across Groupset (AGS) solvers that should be generic to all solvers (even diffusion). This allows me to continue my research on elaborate iterative schemes.
- Added a base Within Groupset (WGS) solver that can be specialized, using a WGS-context, depending on the form of the transport operator. For example, the discrete ordinates solvers has a `SweepWGSContext` which applies sweeping to the inversion of the transport operator.


Naktakala:
Next PR also supersedes this one

PR #310, opened 2023-02-26
Code renaming members 

Naktakala:
# Code-base renaming of members_

This is a VERY large PR where we renamed 99% of all class members according
to our newly adopted convention. The convention is now as follows:

## More on Class member conventions
- All class members should be named the same as regular variables, but with
  a trailing underscore ('_'), e.g., `cell.centroid_`.
- `struct` members should NOT be followed by an underscore.
- Strongly consider making all class members `private/protected` unless
  - They are `const`.
  - They cause a sensitivity to speed, e.g., `chi_mesh::cell` does not have
    accessor functions because `cell.vertex_ids_` is faster than 
    `cell.GetVertexIDs()`
- References to `private/protected` members should be obtained with a method
  starting with `Get`, e.g., `object.member` should be referenced with
  `object.GetMember()`
- Use `const` and non-`const` getters judicially. Favor using setters over
  non-const reference if possible, e.g., setting a single value can be done
  with `GetSingleValue`/`SetSingleValue`, however, vector manipulation might
  be better done with `GetVector` followed by manipulations.


zachhardy (dismissed):
Two things: 1) getting failed regression tests (working on figuring these out, 2) the `object.GetMember()` notation is not followed by the changes in `lbs::LBSSolver`.

I'm fine with not requiring the `object.GetMember()` notation, in general. I think things like `solver.NumGroups()` is clear enough.

zachhardy (approved):

Naktakala:
On 2) The coding conventions have "should", not "must". However, to make it more clear, the doc has been changed to:
- References to `private/protected` members should be obtained with a method
  call. A recommended convention is starting the method call with `Get`, e.g.,
  `object.member` should be referenced with
  `object.GetMember()`. This is only guidance. It is perfectly fine to have
  `object.Member()` or anything that makes sense for the given use case.

Issue #313, opened 2023-02-27
Develop a formulation for better handling of adjoint cross sections


zachhardy:
Currently, the adjoint solver lugs around the transpose of the transfer and production matrices within `matid_to_S_transpose` and `matid_to_F_transpose`. This is done instead of flipping indices for performance reasons. Ideally, there is some more natural accessor, such as a derived `TransportCrossSections` class `AdjointTransportCrossSections`, containing the necessary transposed data, or the cross sections are modified within `TransportCrossSections` with a flag to determine whether or not modifications have been made. The favored method was to leave `TransportCrossSections` alone.

ragusa:
When one loads a XS in chi-tech, one has an object XSobj_forward.

Can we create a derived class that 
- takes in XSobj_forward, make it adjoint, which can mean transposing but maybe also reversing group ordering (**think about that one; it may make re-using LBS easier**)
- and returns to lua XSobj_adjoint. Then, it is the user resposability to pass the proper lua object tot he adjoint solver.

@Naktakala @zachhardy Thoughts?

zachhardy:
@ragusa This is what I am thinking. We can create an object `AdjointTransportCrossSections`. Then we have two options:

1. Make this a derived type of `TransportCrossSections` and use a few initialization methods. First, we can initialize via an instance of `TransportCrossSections` where we copy all vector cross section data and transpose the transfer and production matrices. Second, we can initialize via a .cxs file where the data will be transposed when read. Alternatively, we can have an adjoint .cxs file read in which has the data pre-transposed. 
2. We can create a struct `AdjointTransportCrossSections` which stores only the transposed transfer and production matrices. This can be initialized in the same ways as in the previous item. From this, we avoid copies of all of the other cross section data. During the `SetSource` routine we then query the forward cross sections and the adjoint cross sections.

ragusa:
@zachhardy 
Also, do not overlook the possiblity of transposing  AND reversing the groups (last row becomes first row and so on). Imagine you have a pure down scatterer. If you transpose, the transfer matrix is now upper triangular. It would best to solver from the most thermal group upwards to the first fast group. 

- Maybe this reversing of the group solve is obvious from lua. 
- But even if so, you need to think about the WGS and AGS or the group sets in general. I just have a feeling that it may be worthwhile to also reverse so that the loops so that LBS remains as before, more or less.

zachhardy:
@ragusa 
[...]

PR #315, opened 2023-03-03
Large Scale XS Changes: Accessors and Adjoints

zachhardy:
This PR implements a large scale cross section format change. Now, all cross sections are accessed via accessor functions and adjoint cross sections are implemented. This addresses #313.

### Cross Section Changes

- Created abstract base class `MultiGroupXSBase` 
    - No data stored in the class   
    - Contains virtual accessor functions. All accessors return const references to the requested data.
    - Contains const routines that can be carried out with only calls to accessors
    - Changed fission scaling routine. Now, a `fission_scaling_factor_` is stored. This is defaulted to unity and can be changed via a call to `SetFissionScalingFactor`, which undoes the previous scaling and rescales according to the argument. This is done because the accessor functions return references to the vector data. If the scaling were done upon a query, this would require us to both return a copy of the underlying data *and* require us to iterate over all the data to scale it on *each* query. The implemented method should be significantly more efficient and allows for us to be able to revert the scaling, if necessary, which did not exist in the previous iteration of fission scaling.
   
- Created `MultiGroupXS` class deriving from `MultiGroupXSBase`
    - Stores the cross section data
    - Implements accessor functions and various other routines

- Created `AdjointMultiGroupXS` class deriving from `MultiGroupXSBase`
    - Stores a const reference to a `MultiGroupXS` instance and storage for transposed transfer and production matrices
    - Transposes transfer and production matrices at construction
    - Accesses non-transposed data via reference to `MultiGroupXS`

### Adjoint Solver Changes

- Removed direct storage of transposed transfer and production matrices.
- At initialization, create a new `std::pair<int, std::shared_ptr<MultiGroupXSBase>> matid_to_xs_map` which points to `AdjointMultiGroupXS` pointers and reassign these to the transport cell views.

### General LBS Changes

- All cross section references changed to the accessor functions
- More member_ renaming missed from #310


framework/ChiPhysics/PhysicsMaterial/MultiGroupXS/adjoint_multigroup_xs.h
```diff
@@ -0,0 +1,90 @@
...
+
+#include "multigroup_xs.h"
+
+
+namespace chi_physics
+{
+
+/**
+ * A class for accessing multi-group cross sections for adjoint problems.
+ */
```
Naktakala: We can maybe say a little bit more here. 
The class avoids having to store duplicate data since the only thing that changes is that the transfer- and production matrices are transposed. Therefore we store a reference to the XS this is based off and just reference that data if it is not changed.

framework/ChiPhysics/PhysicsMaterial/MultiGroupXS/multigroup_xs.h
```diff
@@ -0,0 +1,146 @@
...
+/**\defgroup LuaTransportXSs Transport Cross Sections
+ * \ingroup LuaPhysics*/
+
+namespace chi_physics
+{
+
+
+//######################################################################
+/**
+ * A class for handling forward multi-group cross sections.
```
Naktakala: Remove the word "forward" please. This should just be classical multi-group cross sections.

framework/chi_runtime.h
```diff
@@ -31,12 +31,12 @@ namespace chi_physics
 {
   class Solver;
   class Material;
-  class TransportCrossSections;
+  class MultiGroupXS;
```
Naktakala, line 34: This should be changed to `MultiGroupXSBase`

framework/chi_runtime.h
```diff
@@ -31,12 +31,12 @@ namespace chi_physics
...
   class FieldFunction;
 
   typedef std::shared_ptr<Solver>                 SolverPtr;
   typedef std::shared_ptr<Material>               MaterialPtr;
-  typedef std::shared_ptr<TransportCrossSections> TransportCrossSectionsPtr;
+  typedef std::shared_ptr<MultiGroupXS>           MultiGroupXSPtr;
```
Naktakala, line 39: Ditto

modules/DiffusionSolver/Solver/general.cc
```diff
@@ -126,17 +127,14 @@ void chi_diffusion::Solver::GetMaterialProperties(const chi_mesh::Cell& cell,
     bool transportxs_found = false;
     for (int p=0; p<material->properties_.size(); p++)
     {
-      if (std::dynamic_pointer_cast<chi_physics::TransportCrossSections>
+      if (std::dynamic_pointer_cast<chi_physics::MultiGroupXS>
```
Naktakala, line 128: There is no need for the diffusion solver to deal with `MultiGroupXS` is there? We could change it to the base.

modules/LinearBoltzmannSolvers/A_LBSSolver/lbs_structs.h
```diff
@@ -2,7 +2,7 @@
 #define LBS_STRUCTS_H
 
 #include "ChiMath/chi_math.h"
-#include "ChiPhysics/PhysicsMaterial/transportxsections/material_property_transportxsections.h"
+#include "ChiPhysics/PhysicsMaterial/MultiGroupXS/multigroup_xs.h"
```
Naktakala, line 5: So I am expecting this file to only be included in rare circumstances. I can think of only these right now:
- When it is created (lua wrappers), some SimTests
- When an adjoint is created (the .cc in LBS where this happens)

The unfortunate thing here is that the declaration of `class MultiGroupXS` does require that header so we are getting away with it.

The ones in green here need to be changed to the base:
image

Naktakala (changes requested):
Its just some references to the derived class that needs to change to base. And two pieces of documentation. Overall great job!

Naktakala (approved):

ragusa:
Sorry I am slow. Questions:


  1.  I see this in MultiGroupXS/multigroup_xs_pushlua.cc<https://github.com/chi-tech/chi-tech/pull/315/files#diff-2891fe6c22ea503fb6f3ef0de9e15acd375106cda000a3115216417f3fc78af4>:

  //================================================== Push diffusion quantities

  Push1DXS(DiffusionCoefficient(), "diffusion_coeff");

  Push1DXS(SigmaRemoval(), "sigma_removal");

  Push1DXS(SigmaSGtoG(), "sigma_s_gtog");



Name is strange. Why gtog and not gtogp? Gtog feels like the diagonal to me.



  1.  The headers in single_state_mgxs.h<https://github.com/chi-tech/chi-tech/pull/315/files#diff-4e3bb81a80cb255ac783ed5d07588a1dd81cfcebe30c90b5de838ee54b1c8ba2> should be change so as to not to say TRANSPORT_CROSS_SECTIONS



  1.  lua/physics_materialproperties.cc<https://github.com/chi-tech/chi-tech/pull/315/files#diff-0e21b7e7befd6ff502894378a94ac282a2d7beb1e674365c90729aba8790b7f5> stll has transportxs, i.e., using XS = chi_physics::TransportCrossSections;

PR #318, opened 2023-03-07
LBS BC upgrades

Naktakala:
Now supports advanced BCs. This is necessary in preparation for supporting preassigned boundary ids.


framework/ChiMesh/SweepUtilities/AngleSet/angleset.cc
```diff
@@ -161,12 +161,13 @@ PsiBndry(uint64_t bndry_map,
...
-      angle_num, cell_local_id, face_num, fi, gs_ss_begin);
+      cell_local_id, face_num, fi, angle_num, g, gs_ss_begin);
 
   if (not surface_source_active)
-    return &ref_boundaries[bndry_map]->zero_boundary_flux[g];
+    return ref_boundaries[bndry_map]->ZeroFlux(g);
 
-  return &ref_boundaries[bndry_map]->boundary_flux[g];
+  return ref_boundaries[bndry_map]->HeterogenousPsiIncoming(
+    cell_local_id, face_num, fi, angle_num, g, gs_ss_begin);
```
zachhardy, line 170: I understand we use the `zero_boundary_flux` because we return a pointer to a double, but is there a better way to do this without having a to have zero vector?

framework/ChiMesh/SweepUtilities/SweepBoundary/boundary_base.cc
```diff
@@ -0,0 +1,37 @@
...
+                        int fi,
+                        int angle_num,
+                        int gs_ss_begin)
+{
+  chi::log.LogAllError()
+    << "HeterogenousPsiOutgoing call made to boundary "
+       "that has no such information.";
+  chi::Exit(EXIT_FAILURE);
+  return nullptr;
+}
```
zachhardy: I was told not to use Base...

framework/ChiMesh/SweepUtilities/SweepBoundary/sweep_boundaries.h
```diff
@@ -11,97 +12,122 @@ namespace chi_mesh::sweep_management
...
 enum class BoundaryType
 {
-  VACUUM                = 0,
-  INCIDENT_HOMOGENOUS   = 1,
-  REFLECTING            = 2,
-  INCIDENT_HETEROGENOUS = 3
+  INCIDENT_VACCUUM                  = 0, ///< Zero for all angles, space
+  INCIDENT_ISOTROPIC_HOMOGENOUS     = 1, ///< One value for all angles, homogenous in space
+  REFLECTING                        = 2, ///< Reflecting boundary condition about a normal
+  INCIDENT_ANISOTROPIC_HETEROGENOUS = 3  ///< Complex different for each angle and face node
```
zachhardy, line 18: What i I have a space-dependent isotropic boundary condition or a space-independent anisotropic boundary condition?

framework/ChiMesh/SweepUtilities/SweepBoundary/sweep_boundaries.h
```diff
@@ -11,97 +12,122 @@ namespace chi_mesh::sweep_management
...
+  std::vector<double> boundary_flux_;
 public:
   explicit
-  BoundaryVacuum(const std::vector<double>& ref_boundary_flux) :
-  BoundaryBase(BoundaryType::VACUUM,ref_boundary_flux)
+  BoundaryVaccuum(size_t in_num_groups,
+                  chi_math::CoordinateSystemType coord_type =
+                    chi_math::CoordinateSystemType::CARTESIAN) :
+    BoundaryBase(BoundaryType::INCIDENT_VACCUUM, in_num_groups, coord_type),
+    boundary_flux_(in_num_groups, 0.0)
```
zachhardy: See my previous comment on `zero_boundary_flux`.

framework/ChiMesh/SweepUtilities/SweepBoundary/sweep_boundaries.h
```diff
@@ -110,39 +136,117 @@ class BoundaryReflecting : public BoundaryBase
...
+    const std::vector<std::pair<double,double>>& quadrature_phi_theta_angles,
+    const std::vector<int>& group_indices)
+  {
+    size_t num_angles = quadrature_angle_indices.size();
+    size_t num_groups = group_indices.size();
+
+    std::vector<double> psi(num_angles * num_groups, 0.0);
+
+    return psi;
+  }
```
zachhardy: We should add a default argument `const double time = 0.0` to this to handle transients.


Is there a reason for returning zeros? Why not make this abstract?

modules/LinearBoltzmannSolvers/A_LBSSolver/Tools/lbs_bndry_func_lua.h
```diff
@@ -0,0 +1,36 @@
...
+    size_t        cell_global_id,
+    int           cell_material_id,
+    unsigned int  face_index,
+    unsigned int  face_node_index,
+    const chi_mesh::Vector3& face_node_location,
+    const chi_mesh::Vector3& face_node_normal,
+    const std::vector<int>& quadrature_angle_indices,
+    const std::vector<chi_mesh::Vector3>& quadrature_angle_vectors,
+    const std::vector<std::pair<double,double>>& quadrature_phi_theta_angles,
+    const std::vector<int>& group_indices) override;
```
zachhardy: We should add a default argument `const double time = 0.0` to this to handle transients.

modules/LinearBoltzmannSolvers/A_LBSSolver/lua/lbs_lua_utils.cc
```diff
@@ -52,6 +52,7 @@ namespace lbs::common_lua_utils
     RegisterNumberValueToTable(VACUUM            ,1,LBSBoundaryTypes);
     RegisterNumberValueToTable(INCIDENT_ISOTROPIC,2,LBSBoundaryTypes);
     RegisterNumberValueToTable(REFLECTING        ,3,LBSBoundaryTypes);
+    RegisterNumberValueToTable(INCIDENT_ANISTROPIC_HETEROGENOUS,4,LBSBoundaryTypes);
```
zachhardy, line 55: See previous comment on these labels.

zachhardy (approved):
Feel free to address any or all of my comments. I would recommend against `BoundaryBase` as that a wise person once told me to avoid using Base in the name...

framework/ChiMesh/SweepUtilities/SweepBoundary/boundary_iso_homo.cc
```diff
@@ -0,0 +1,14 @@
+#include "sweep_boundaries.h"
+
+//###################################################################
+/**Returns a pointer to a heterogenous flux storage location.*/
```
ragusa: the filename has homo which I interpret as homogenous but the comment says heterogenous. I am confused.

framework/ChiMesh/SweepUtilities/SweepBoundary/boundary_vaccuum.cc
```diff
@@ -0,0 +1,14 @@
+#include "sweep_boundaries.h"
+
+//###################################################################
+/**Returns a pointer to a heterogenous flux storage location.*/
```
ragusa, line 4: same about the comment and the filename

tests/Transport_Steady/Transport2D_5PolyA_AniHeteroBndry.lua
```diff
@@ -0,0 +1,190 @@
+-- 2D Transport test with Vacuum and Incident-isotropic BC.
```
ragusa, line 1: thanks for adding a test! I can't wait to use this soon!

framework/ChiMesh/SweepUtilities/AngleAggregation/angleaggregation.cc
```diff
@@ -49,8 +60,8 @@ void chi_mesh::sweep_management::AngleAggregation::ZeroIncomingDelayedPsi()
     {
       auto& rbndry = (BoundaryReflecting&)(*bndry);
 
-      if (rbndry.opposing_reflected)
-        for (auto& angle : rbndry.hetero_boundary_flux_old)
+      if (rbndry.IsOpposingReflected())
```
ragusa, line 63: I looked down two lines below what was shown in the github diff (that is, I looked at line 69) and I do not understand the val = 0. What are those nested loops actually doing?

ragusa (approved):
a few of comments, if you could reply before merging so that I learn something!

framework/ChiMesh/SweepUtilities/SweepBoundary/boundary_base.cc
Naktakala: Very good point. Will change it

framework/ChiMesh/SweepUtilities/SweepBoundary/sweep_boundaries.h
Naktakala: I don't know what you have. Basically the `INCIDENT_ANISOTROPIC_HETEROGENOUS` can be used to emulate both `INCIDENT_VACCUUM` and `INCIDENT_ISOTROPIC_HOMOGENOUS`.

framework/ChiMesh/SweepUtilities/AngleAggregation/angleaggregation.cc
Naktakala: If the simulation has opposing reflected boundary conditions we essentially have a cycle there. One of the boundaries has to break the cycle, store delayed angular fluxes for 
- a particular angle,
- a particular cell,
- a particular face or faces on that cell (not all),
- a set of DOFs

and these need to be iterated on. Because the offsets here can be so irregular, we are just using a N-dimensional array. I am open to ideas if you can think of doing this in a simpler manner.

tests/Transport_Steady/Transport2D_5PolyA_AniHeteroBndry.lua
Naktakala: :-)

framework/ChiMesh/SweepUtilities/SweepBoundary/boundary_iso_homo.cc
Naktakala: Oops. Great catch. Probably copy paste error

framework/ChiMesh/SweepUtilities/SweepBoundary/boundary_vaccuum.cc
Naktakala: Will get it fixed.

framework/ChiMesh/SweepUtilities/AngleSet/angleset.cc
Naktakala: All ears if you find a better way whilst maintaining the cache behavior within the sweep.

framework/ChiMesh/SweepUtilities/SweepBoundary/sweep_boundaries.h
Naktakala: Two very good points

modules/LinearBoltzmannSolvers/A_LBSSolver/Tools/lbs_bndry_func_lua.h
Naktakala: Ditto

framework/ChiMesh/SweepUtilities/SweepBoundary/sweep_boundaries.h
zachhardy: This comment was mainly to highlight that under this option set, you have to have angle-space independence or angle-space dependence, but one dependent and one independent. I understand that with `INCIDENT_ANISOTROPIC_HETEROGENEOUS` you can specify anything.

Naktakala: Not sure I understand the point here?

PR #323, opened 2023-03-13
New set source function classes

Naktakala:
In an effort to reduce code duplication a `SourceFunction` has been created with separate pieces that can be customized by different solvers.


zachhardy (approved):
Nice

PR #335, opened 2023-05-22
Big change following KEigen upgrades

Naktakala:
This is a terribly big PR. It is the brunt of the implementation of static registration.

I wouldn't expect anyone to review this.


zachhardy (approved):
Your expectation is valid :) As long as the regression tests pass, it will be easier to correct any issues discovered late with small PRs than to attempt to review this... I know the gist anyways, so good to go for me

PR #338, opened 2023-06-12
Improvements 2023 06 w2

Naktakala:
## Highlights
- Fixed a bug with `ParameterBlock::ChangeToArray` where it only applied
  the change if the array was empty.
- Added a base class `MeshModifier` and the first implementation 
  `SnapToPlaneMeshModifier`. This will serve as a basis to move away from using
  `chiVolumeMesherSetProperty`
- Changed `LogicalVolume` to a `ChiObject` and moved it to the 
  `chi::object_stack`
- Fixes for reading from VTK with the ability to specify the field from which
  material IDs are read


framework/ChiMesh/FieldFunctionInterpolation/Volume/chi_ffinter_volume.h
```diff
@@ -2,7 +2,7 @@
 #define CHI_FFINTER_VOLUME_H
 
 #include "../chi_ffinterpolation.h"
-#include "ChiMesh/LogicalVolume/chi_mesh_logicalvolume.h"
+#include "ChiMesh/LogicalVolume/LogicalVolume.h"
```
zachhardy, line 5: Why the sudden change in file name syntax?

framework/ChiMesh/FieldFunctionInterpolation/lua/ffinterpol_setproperty.cc
```diff
@@ -335,7 +335,8 @@ int chiFFInterpolationSetProperty(lua_State* L)
     int logvol_hndle = lua_tonumber(L, 3);
 
     auto p_logical_volume =
-      chi::GetStackItemPtr(chi::logicvolume_stack, logvol_hndle, fname);
+      std::dynamic_pointer_cast<chi_mesh::LogicalVolume>(
+      chi::GetStackItemPtr(chi::object_stack, logvol_hndle, fname));
```
zachhardy, line 339: Is everything going to be migrated to the object stack, or just some things?

framework/ChiMesh/LogicalVolume/BooleanLogicalVolume.cc
```diff
@@ -0,0 +1,58 @@
...
+  const auto& input_parts = params.GetParam("parts");
+  input_parts.RequireBlockTypeIs(chi_objects::ParameterBlockType::ARRAY);
+
+  for (size_t p = 0; p < input_parts.NumParameters(); ++p)
+  {
+    const auto& part = input_parts.GetParam(p);
+    part.RequireBlockTypeIs(chi_objects::ParameterBlockType::BLOCK);
+
+    part.RequireParameter("op");
+    part.RequireParameter("lv");
```
zachhardy: These should be either obvious from the name or documented in a syntax block in my view.

framework/ChiMesh/LogicalVolume/BooleanLogicalVolume.h
```diff
@@ -0,0 +1,24 @@
...
+
+namespace chi_mesh
+{
+
+// ###################################################################
+/**Boolean volume*/
+class BooleanLogicalVolume : public LogicalVolume
+{
+public:
+  std::vector<std::pair<bool, std::shared_ptr<const LogicalVolume>>> parts;
```
zachhardy, line 14: Does the bool imply within or not within a part?

framework/ChiMesh/LogicalVolume/RCCLogicalVolume.cc
```diff
@@ -0,0 +1,94 @@
...
+  params.AddOptionalParameter("r", 1.0, "Radius of the sphere.");
+  params.AddOptionalParameter("x0", 0.0, "X-coordinate of the volume base");
+  params.AddOptionalParameter("y0", 0.0, "Y-coordinate of the volume base");
+  params.AddOptionalParameter("z0", 0.0, "Z-coordinate of the volume base");
+  params.AddOptionalParameter(
+    "vx", 0.0, "X-component of the volume extrusion vector");
+  params.AddOptionalParameter(
+    "vy", 0.0, "Y-component of the volume extrusion vector");
+  params.AddOptionalParameter(
+    "vz", 1.0, "Z-component of the volume extrusion vector");
```
zachhardy, line 29: Interestingly, in most other places the clang-format is turned off at the parameter specification block and not here. Is this necessary? What's so bad about an indent?

framework/ChiMesh/LogicalVolume/RPPLogicalVolume.cc
```diff
@@ -0,0 +1,59 @@
...
+  params.AddOptionalParameter("ymax", 1.0, "Y-max of the volume");
+  params.AddOptionalParameter("zmin", 0.0, "Z-min of the volume");
+  params.AddOptionalParameter("zmax", 1.0, "Z-max of the volume");
+
+  params.AddOptionalParameter(
+    "infx", false, "Flag, when true, will ignore xmin and xmax.");
+  params.AddOptionalParameter(
+    "infy", false, "Flag, when true, will ignore ymin and ymax.");
+  params.AddOptionalParameter(
+    "infz", false, "Flag, when true, will ignore zmin and zmax.");
```
zachhardy, line 32: This is much better than the old way of just specifying big numbers

framework/ChiMesh/LogicalVolume/SurfaceMeshLogicalVolume.cc
```diff
@@ -0,0 +1,213 @@
...
+SurfaceMeshLogicalVolume::SurfaceMeshLogicalVolume(
+  const chi_objects::InputParameters& params)
+  : LogicalVolume(params),
+    surf_mesh(chi::GetStackItemPtrAsType<chi_mesh::SurfaceMesh>(
+      chi::object_stack,
+      params.GetParamValue<size_t>("surface_mesh_handle"),
+      __FUNCTION__)),
+    xbounds_({1.0e6, -1.0e6}),
+    ybounds_({1.0e6, -1.0e6}),
+    zbounds_({1.0e6, -1.0e6})
```
zachhardy, line 42: Does this high, then low specification make sense? Usually you see low, then high (i.e. increasing order).

framework/ChiMesh/LogicalVolume/SurfaceMeshLogicalVolume.cc
```diff
@@ -0,0 +1,213 @@
...
+    {
+      xbounds_[0] = std::min(xbounds_[0],x);
+      xbounds_[1] = std::max(xbounds_[1],x);
+      ybounds_[0] = std::min(ybounds_[0],y);
+      ybounds_[1] = std::max(ybounds_[1],y);
+      zbounds_[0] = std::min(zbounds_[0],z);
+      zbounds_[1] = std::max(zbounds_[1],z);
+    }
+  }
+}
```
zachhardy, line 69: If I am reading this correctly, this is finding the mesh bounds from a collection of vertices. On my previous comment... You are setting xbounds_[0] with a min and xbounds_[1] with a max, whereas in the constructor it was specified as max, then min.

framework/chi_runtime.cc
```diff
@@ -213,7 +212,7 @@ void chi::Finalize()
   meshhandler_stack.clear();
 
   surface_mesh_stack.clear();
-  logicvolume_stack.clear();
```
zachhardy, line 216: `object_stack` is cleared below. This line should be able to be omitted.

zachhardy (changes requested):
A few small comments and requests. Leaving the VTK stuff alone because I am not very familiar with it.

framework/ChiMesh/FieldFunctionInterpolation/Volume/chi_ffinter_volume.h
Naktakala: I am giving it a go. Saw it in some NRC codes. Basically, when a class' declaration and definitions can stay as a pair then we can keep it CamelCase, however, in order to hint that a class's definition is split among multiple files the syntax changes to snake_case. The lower case is anyhow a "should", not a must ([see doc standards](https://chi-tech.github.io/db/d7b/_dev_man_code_conventions.html). I am thinking that we can start to adopt this.

framework/ChiMesh/FieldFunctionInterpolation/lua/ffinterpol_setproperty.cc
Naktakala: I would very much like to, however, its not necessarily going to be easy and fast. I'm just one person doing it. Luckily, up to now, I have been able to keep it behind the scenes. Not affecting other code too much.

framework/ChiMesh/LogicalVolume/BooleanLogicalVolume.cc
Naktakala: Agreed.

framework/ChiMesh/LogicalVolume/BooleanLogicalVolume.h
Naktakala: Within. Ties with better documentation. Will get it done.

framework/ChiMesh/LogicalVolume/RCCLogicalVolume.cc
Naktakala: I want to actually see this happening in all code using `SetGeneralDescription`. It seems clang-format is not real consistent with the formatting and in some cases can scrunch up all the code to the right or split sentence so many times that it becomes hard to read. In these sections I want developers to have the full space available.

To answer the second question... having an indent here wouldn't be bad, but on the same note NOT having one wouldn't either.

framework/ChiMesh/LogicalVolume/SurfaceMeshLogicalVolume.cc
Naktakala: The way this has been recoded these values aren't used, but it still requires something. In that sense:
`xmin,xmax = {-12, +12}`
first value `5`
`xmin = min(5, -12) = -12` and
`xmax = max(5, +12) = +12`. Clearly wrong.

Versus the other way around:
`xmin, xmax = {+12, -12}`
first value `5`
`xmin = min(5, +12) = 5` and
`xmax = max(5, -12) = 5`
works.

framework/ChiMesh/LogicalVolume/SurfaceMeshLogicalVolume.cc
Naktakala: Given my response to the previous comment I believe this will make more sense now. With this code the values in the initializer list dont matter anymore. aka more elegant IMHO

framework/chi_runtime.cc
Naktakala: So here it shows up as red for me. Meaning it was indeed removed. Does it show different for you?

framework/ChiMesh/LogicalVolume/RCCLogicalVolume.cc
zachhardy: Maybe it is just my settings, but when you do the "Refactor" on a block of code within the clang-format off block, does it change? Mine does

framework/ChiMesh/LogicalVolume/SurfaceMeshLogicalVolume.cc
zachhardy: Very sneaky

framework/chi_runtime.cc
zachhardy: It shows as red for me here right now, but before it showed a green line on the other side that said `object_stack.clear();`. Weird. All good.

framework/ChiMesh/LogicalVolume/RCCLogicalVolume.cc
Naktakala: Mine does not! Interesting that yours doesn't

framework/ChiMesh/LogicalVolume/BooleanLogicalVolume.cc
Naktakala: Done.

framework/ChiMesh/LogicalVolume/BooleanLogicalVolume.h
Naktakala: Done

zachhardy (approved):

PR #355, opened 2023-08-07
Update 2023 08 week1

Naktakala:
## Highlights
- Minor exodus exported tweak
- Enhancement to `allowable_range`
- Test system: added weight class. Tests can now be marked "short" or "long"
- Updated workflow for class01 to run more tests

### DIFF
--- a/.github/workflows/Regression.yaml
+++ b/.github/workflows/Regression.yaml
@@ -7,7 +7,7 @@ on:
     
 jobs:
   test:
-    runs-on: 'self-hosted'
+    runs-on: '[self-hosted, class01]'
     steps:
       - uses: actions/checkout@v2
       - name: configure
@@ -27,4 +27,4 @@ jobs:
         run: |
           export MODULEPATH=/scratch-local/software/modulefiles
           module load chitech/gcc/12.3.0
-          test/run_tests -d test/modules
+          test/run_tests -d test/ -j 32 -v 1 
--- a/.github/workflows/Regression.yaml
+++ b/.github/workflows/Regression.yaml
@@ -27,4 +27,4 @@ jobs:
         run: |
           export MODULEPATH=/scratch-local/software/modulefiles
           module load chitech/gcc/12.3.0
-          test/run_tests -d test/ -j 32 -v 1 
+          test/run_tests -d test/ -j 32 -v 1 -w 7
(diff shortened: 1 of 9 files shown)


.github/workflows/Regression.yaml
```diff
@@ -27,4 +27,4 @@ jobs:
         run: |
           export MODULEPATH=/scratch-local/software/modulefiles
           module load chitech/gcc/12.3.0
-          test/run_tests -d test/modules
+          test/run_tests -d test/ -j 32 -v 1
```
wdhawkins: 👍

doc/PAGES/ProgrammersManual/devman_01_07_TestSystem.h
```diff
@@ -94,6 +94,8 @@ Specifies a specific test
 - `"num_procs"` : The number of mpi processes to use.
 - `"checks"` : An array of checks `[..Checks..]`
 - `"args"` : An array of arguments to pass to the executable
+- `"weight_class"` : An optional string, either "short" or "long" used to filter different
+   tests length. The default is "short". "long" should be used for tests >15s.
```
wdhawkins, line 98: 15s seems short to me.  Maybe 5 min?  Or if that seems too long, maybe a couple of minutes? When I think of long running tests, I really think of something that takes an hour or more.

wdhawkins (approved):

doc/PAGES/ProgrammersManual/devman_01_07_TestSystem.h
Naktakala: How about "short" (<15s), "intermediate" (<5min), "long" (>5min). 

I've taken the 15s from MOOSE. I can see why they do it, anything longer and you are unsure if "is it just me or is this thing running long?". I've experienced this myself. The "short" is a nice place to put all the unit tests and such. They are the first line of defense and you don't want them mixed into the longer running tests. On that note, we can make that 30s too.

"short", "intermediate" and "long"?

Naktakala: @ragusa Remember you querying the 3 min run time on the DSA test?

wdhawkins: I think "short", "intermediate", and "long" sounds like a good solution.

PR #356, opened 2023-08-08
Hotfix adjoint tgdsa 2023 08

Naktakala:
`Initialize` routine made to mimic parent class `DiscreteOrdinates`. This was done such that the transposed cross-sections can be used to init DSA routines.


framework/mesh/MeshContinuum/chi_meshcontinuum_exportexodus.cc
```diff
@@ -153,7 +153,8 @@ void chi_mesh::MeshContinuum::
       //then exodus would create the original triangle. Eventually I found that
       //there was an offset of +2 after numerous paraview visualizations and
```
ragusa, line 154: is the comment of +2 still valid? should it be +1? and why is that ????

ragusa (approved):

PR #361, opened 2023-08-10
Update LICENSE to reflect span of years

ragusa:


LICENSE
```diff
@@ -1,6 +1,6 @@
 MIT License
 
-Copyright (c) 2019 The Texas A&M University System, an agency of the State of Texas and chi-tech contributors
+Copyright (c) 2019-2023 The Texas A&M University System, an agency of the State of Texas and chi-tech contributors
```
Naktakala, line 3: We looked at this a while ago. I don't know what this really means and there is not a whole lot of examples of other codes updating this number year on year. Examples:  
Hypre: `Copyright (c) 1998 Lawrence Livermore National Security, LLC and other`
PETSc: `Copyright (c) 1991-2021, UChicago Argonne, LLC and the PETSc Development Team`
deal.ii: All over the place `1991, 1999` and `2004, 2006`
GEANT4: `Copyright (c) Copyright Holders of the Geant4 Collaboration, 1994-2006.`

OpenMC and OpenMOC seems to keep it up to date.

I am open for a discussion.

Naktakala: I would like @wdarylhawkins to comment on this. I see MOOSE just avoids the issue entirely by using GNU v2.1

Naktakala: Another question: If we have to update this year to year, does it mean that if we forget to do this January 1, 2024 and only get to it February 1, 2024 that our copyright disappears for a month (making it a free-for-all for a month)???

My opinion is to put years on there every time the license is changed. e.g. 2019 for now, if we add stuff for LANL then 2019, 2023.

ragusa: I tend to want to follow closely what OpenMC does. They seem to religiously update it every year. 

Is there a specific reason you think it should not be updated in years when changes are being made? 

According to https://softwareengineering.stackexchange.com/questions/210472/is-renewal-of-mit-license-needed-on-github-at-the-beginning-of-each-year, 

"The copyright notice should include the year in which you finished preparing the release (so if you finished it in 1998 but didn't post it until 1999, use 1998). You should add the proper year for each release; for example, “Copyright 1998, 1999 Terry Jones” if some versions were finished in 1998 and some were finished in 1999. If several people helped write the code, use all their names.

For software with several releases over multiple years, it's okay to use a range (“2008-2010”) instead of listing individual years (“2008, 2009, 2010”) if and only if every year in the range, inclusive, really is a “copyrightable” year that would be listed individually; and you make an explicit statement in your documentation about this usage."

The article concludes with:
-if you made changes that year, do include the year in a comma-separated list in your copyright notice.
-if did not make copyrightable changes that year, do not include that year in your copyright notice.

Thus, since we are still changing files, it is more proper to have 2019-2023 and to update every year.

Naktakala: I am really impartial on this. Was just worried that we'd forget and what the implications would be. But again, I don't really know what those year statements really mean anyways. 

I've set an annual reminder in Slack for this now. 

If @wdarylhawkins has nothing to comment by tomorrow we can merge!

ragusa:
As long as LANL does not control of the code maintenance, their contributions should be included in the “chi tech contributors” and that’s all.

wdhawkins:
Sorry for the late response.  I saw this yesterday afternoon and it slipped off my radar.  I'm going to preface everything I say here with the usual "I'm not a lawyer." statement. But, here are my thoughts and what I remember from going through this with PDT.  The MIT license is fine. The copyright notice should absolutely include the first year the code was written.  After that, other years in which development occurred should also be added, but forgetting a year or failing to update the license does not invalidate the copyright. LANL contributions to the code can be acknowledged by simply adding "Los Alamos National Laboratory (LANL)" to the copyright statement. The organizational acknowledgement will cover everyone from LANL that contributes without having to list them individually (so long as they contributed as part of their LANL duties).

Naktakala (approved):

PR #362, opened 2023-08-16
Update 2023 08 week3

Naktakala:
## Highlights
- **New `MeshGenerator` system**. This system will be replacing the 
  `SurfaceMesher`/`VolumeMesher` pairs and provide a much simpler interface to 
  users for creating meshes. Additionally, this will be waaaay more extendable
  than what we have now. We have the following mesh generators:
  - `MeshGenerator`, the base, can be used to link to toghether other generators
  - `OrthogonalMeshGenerator`
  - `FromFileMeshGenerator`, a one-stop shop for reading meshes from files
  - `ExtruderMeshGenerator`, very sleak extruded with lots of options     
  
- **New `Partitioner` system**. We now have a new set of partitioners:
  - `PETScGraphPartitioner`
  - `LinearGraphPartitioner`
  - `KBAGraphPartitioner`
- Some documentation additions.
- Fixed a bug with exporting meshes in parallel via `chiMeshHandlerExportMeshToVTK`
- Fixed a rather nasty bug with `lbs::MakeSubSet`. This method has been removed
  and it's fixed counterpart added under `utils/chi_utils.h`
- Fixed a sorting problem with `ParameterBlock` when used for arrays.


wdhawkins:
Sorry I'm late on this.  I'm on vacation this week and next week.  This all looks good to me.  There was a lot here, but I didn't see anything that just jumped out at me.

wdhawkins (approved):

Naktakala:
Comments from @ragusa:

- What is a “real mesh”?
- What are monotonic nodes ???
- “with another mesh generator as an input…” this sentence is confusing. Maybe you meant to say with another mesh generated in 2D as an input …
-I am unsure about the LinearGraphPartitioner. What does it do? What is a non-simulation mesh??
- KBA: it has different flavors, including hybrid KBA. So if it the original KBA or whatever was in PDT. Not sure we should reference PDT’s as the “classic KBA”.
- In the last figure, did classic KBA cut the cells? I can’t really tell from the picture if there are orthogonal cutlines in xy.

Naktakala:
- Wrt to the "real" mesh. A real mesh is the proper and usable `chi_mesh::MeshContinuum` object with all the bells and whistle's initialized. To make it more clear I removed the "real" and just replaced it with `chi_mesh::MeshContinuum`
- Cannot tell if you didn't know what `monotonic` means or whether you want me to explicitly say `monotonically increasing`. Rearranged to say `monotonically increasing`.
- Hmmm, how to reduce the confusion? The basic concept here is that we no longer need to have the 2D mesh to be a single input. We can, in future, extend the Extruder to extrude 1D meshes, have mixed meshes (1D->2D, 2D->3D), have parts of a 2D mesh from diifferent mesh generators, etc. So the basic example here is the use of another mesh generator as an input.
- The Linear partitioner is useful for a lot of things, including eventually reading meshes directly into partitions. It is useful for simulations where you need proxy cells for visualization or some other utility, but the cells do not have to be connected to each other in a sense that is needed for proper ParMETIS or KBA partitioning. In that same sense a non-simulation mesh is not the mesh that forms the basis for a field (e.g., FEM field) and therefore cannot directly be used for a simulation. To avoid confusion I removed the word "non-simulation" and just replaced it with "a very basic partitioner used for during the preparation of simulation meshes". Also added some wording for the `LinearGraphPartitioner`'s doc-string.
- KBA: Lets start a revolution!? Please join me. Original KBA overlayed a Px times Py orthogonal grid over the mesh and used that for partitioning. People started calling that "KBA". Then, when Pz is slapped on, there seems to be a burning desire to preserve the "KBA is strictly 2D" nostalgia by calling it a different name (i.e., HybridKBA). This is like calling 3D ortho meshes "Hybrid 2D meshes"? In the spirit of helping the user out I am choosing to call the PxPyPz business general "KBA"-style instead of confusing the user by creating `KBAGraphPartitioner` as well as a `HybridKBAPartitioner`, and obviously avoid things like `KBAUnlessSpecifyingPzThenHybridKBAGraphPartitioner`.
- No, KBA partitioning does not automatically cut. We will have to do that with a separate mesh modifier/generator.

wdhawkins:
>     * KBA: Lets start a revolution!? Please join me. Original KBA overlayed a Px times Py orthogonal grid over the mesh and used that for partitioning. People started calling that "KBA". Then, when Pz is slapped on, there seems to be a burning desire to preserve the "KBA is strictly 2D" nostalgia by calling it a different name (i.e., HybridKBA). This is dumb, are we going to call 3D ortho meshes "Hybrid 2D meshes"? In the spirit of helping the user out I am choosing to call the PxPyPz business general "KBA"-style instead of confusing the user by creating `KBAGraphPartitioner` as well as a `HybridKBAPartitioner`, and obviously avoid things like `KBAUnlessSpecifyingPzThenHybridKBAGraphPartitioner`.

I agree with this.  I've never thought of KBA as strictly 2D in processor space. I think of KBA as a columnar spatial decomposition with Pz >= 1. And, as a follow up, there was a time that hybrid KBA meant Pz = 2 (it still does to some).

PR #364, opened 2023-08-28
Parallel vectors

zhardy-lanl:
## Parallel Vector Implementations

This PR introduces the `chi_math::ParallelVector` and `chi_math::GhostedParallelVector` classes. These are primarily intended for use cases when ghosted values of local STL vectors are needed (i.e. PWLD -> PWLC projections) and for non-local contributions to STL vectors.

The `chi_math::ParallelVector` assembly is done in a similar way to PETSc, allowing for either `SET_VALUE` operations or `ADD_VALUE` operations, but not both. Additionally, local vectors can be set using a scalar value or an STL vector. Only local values are available on each process, so all accessors and iterators function only on local values. Lastly, local STL vectors can be generated and returned via a `MakeLocalVector` routine.

The `chi_math::GhostedParallelVector` is derived from `chi_math::ParallelVector`, but additionally pairs the parallel vector with a `chi_math::VectorGhostCommunicator`. The additional functionality in this class is the ability to return a ghosted local vectors, read local or ghost data via global ids, and to update the ghosted values via communication. Apart from read access via global ids, all other accessors are inherited, and therefore only provide access to locally owned data to prevent accidental changes to ghosted data.

A rudimentary unit test is located at `test/framework/chi_math/chi_math_test_02.lua`.


framework/math/ghosted_parallel_vector.h
```diff
@@ -0,0 +1,78 @@
...
+
+
+namespace chi_math
+{
+
+
+class GhostedParallelVector : public ParallelVector
+{
+public:
+  GhostedParallelVector() = delete;
```
Naktakala: Remove this line. Default constructors are defined if and only if it is found that the class has no constructor. Once a constructor is defined the default constructor is not defined and therefore does not need to be deleted.

framework/math/ghosted_parallel_vector.h
```diff
@@ -0,0 +1,78 @@
...
+                            other.comm_)
+  {}
+
+  GhostedParallelVector(GhostedParallelVector&& other)
+    : GhostedParallelVector(other.local_size_,
+                            other.global_size_,
+                            other.ghost_ids_,
+                            other.comm_)
+  {}
+
```
Naktakala: Doc string:
Returns the number of ghost entries that are locally associated with the vector.

framework/math/ghosted_parallel_vector.h
```diff
@@ -0,0 +1,78 @@
...
+  uint64_t NumGhosts() const { return num_ghosts_; }
+  uint64_t TotalLocalSize() const override { return local_size_ + num_ghosts_; }
```
Naktakala: Doc string:
Returns the number of local and ghost entries for the current location.

framework/math/ghosted_parallel_vector.h
```diff
@@ -0,0 +1,78 @@
...
+
+  std::vector<double> MakeGhostedLocalVector() const { return values_; }
```
Naktakala: Doc string:
Returns an STL vector that has the current local values along with the ghost entries appended to the back.

framework/math/parallel_vector.h
```diff
@@ -0,0 +1,140 @@
...
+    : local_size_(other.local_size_),
+      global_size_(other.global_size_),
+      location_id_(other.location_id_),
+      process_count_(other.process_count_),
+      values_(other.values_),
+      comm_(other.comm_)
+  {}
+
+  uint64_t LocalSize() const { return local_size_; }
+  virtual uint64_t TotalLocalSize() const { return local_size_; }
```
Naktakala: This method is horribly out of scope for a non-ghosted vector. Remove it from `ParallelVector`.

framework/math/ghosted_parallel_vector.h
Naktakala: Coupled with my comment in `ParallelVector`, remove the override here.

framework/math/ghosted_parallel_vector.h
```diff
@@ -0,0 +1,78 @@
...
+  /**
+   * Return the value of the parallel vector for the specified global index.
+   *
+   * An error is thrown if the global index does not belong to a locally
+   * owned, or ghost entry.
+   */
+  double GetGlobalValue(const int64_t global_id) const;
+
+  /**
+   * Communicate the current ghost entries to all other processes to
```
Naktakala: Rephrase:
Communicate all current local-entries, that are ghost-entries on other ranks, to their respective recipients. This is a collective call.

framework/math/parallel_vector.cc
```diff
@@ -0,0 +1,264 @@
...
+std::vector<double> ParallelVector::MakeLocalVector()
+{
+  return std::vector<double>(values_.begin(),
+                             values_.begin() + local_size_);
+}
+
+
+double ParallelVector::operator[](const int64_t local_id) const
+{
+  ChiInvalidArgumentIf(
```
Naktakala: `ChiInvalidArgumentIf` is a macro defined as:
```
#define ChiInvalidArgumentIf(condition, message)                               \
  if (condition)                                                               \
  throw std::invalid_argument(std::string(__PRETTY_FUNCTION__) + ": " +        \
                              message)
```
Please remove `std::string(__FUNCTION__)` since the `__PRETTY_FUNCTION__` is already there.

framework/math/parallel_vector.cc
```diff
@@ -0,0 +1,264 @@
...
+      local_id < 0 or local_id >= local_size_,
+      std::string(__FUNCTION__) + ": Invalid local index provided.");
+  return values_[local_id];
+}
+
+
+double& ParallelVector::operator[](const int64_t local_id)
+{
+  ChiInvalidArgumentIf(
```
Naktakala: Ditto

framework/math/parallel_vector.cc
```diff
@@ -0,0 +1,264 @@
...
+void ParallelVector::SetValue(const int64_t global_id,
+                              const double value,
+                              const OperationType op_type)
+{
+  ChiInvalidArgumentIf(
+      global_id < 0 or global_id >= global_size_,
+      "Invalid global index encountered. Global indices "
+      "must be in the range [0, this->GlobalSize()].");
+  if (op_type == OperationType::SET_VALUE)
+    set_cache_.push_back({global_id, value});
```
Naktakala: Use an `emplace_back` here instead. Avoids the copy.

framework/math/parallel_vector.cc
```diff
@@ -0,0 +1,264 @@
...
+  else
+    add_cache_.push_back({global_id, value});
+}
+
+
+void ParallelVector::SetValues(const std::vector<int64_t>& global_ids,
+                               const std::vector<double>& values,
+                               const OperationType op_type)
+{
+  ChiInvalidArgumentIf(
```
Naktakala: Ditto

framework/math/parallel_vector.cc
```diff
@@ -0,0 +1,264 @@
...
+      global_ids.size() != values.size(),
+      std::string(__FUNCTION__) + ": " +
+      "Size mismatch between indices and values.");
+
+  for (size_t i = 0; i < global_ids.size(); ++i)
+    SetValue(global_ids[i], values[i], op_type);
```
Naktakala: We can squeeze a little performance out here. Since we know the optype we can avoid one*N if checks in SetValue.
```
auto& op_cache = op_type == OperationType::SET_VALUE ? set_cache_ : add_cache_;
for (size_t i = 0; i < global_ids.size(); ++i)
{
   ChiInvalidArgumentIf(
      global_id < 0 or global_id >= global_size_,
      "Invalid global index encountered. Global indices "
      "must be in the range [0, this->GlobalSize()].");
  op_cache.emplace_back(global_ids[i], values[i]);
}
```

Naktakala (changes requested):

framework/math/ParallelVector/parallel_vector.h
```diff
@@ -0,0 +1,157 @@
...
+
+  /**
+   * Read/write accessor to the entry at the given local index of
+   * the local vector.
+   *
+   * \note This accessor only allows access to the locally owned entries
+   *       of the parallel vector, and does not allow access for any data
+   *       beyond local_size_ that may exist in derived classes.
+   */
+  double& operator[](const int64_t local_id);
```
Naktakala, line 79: Another problem here. If you use this function in pure read fashion, now all of a sudden you can't read ghosts just because you are using it in a method/function where it is non-const.

framework/math/ParallelVector/parallel_vector.h
```diff
@@ -0,0 +1,157 @@
...
+#include <cstdint>
+#include <map>
+
+#include <mpi.h>
+
+
+namespace chi_math
+{
+
+enum class OperationType : short
```
Naktakala: Forgot about this one previous round. Can you please make this `VecOpType` or something that ties it to vectors. Right now this will land in `chi_math::OperationType` which could easily clash with something in future or be very ambiguous.

framework/math/ParallelVector/parallel_vector.h
```diff
@@ -0,0 +1,157 @@
...
+};
+
+
+/**
+ * An implementation of a parallel vector.
+ */
+class ParallelVector
+{
+public:
+  using iterator = std::vector<double>::iterator;
```
Naktakala: Are these used somewhere now that iterator is gone?

framework/math/ParallelVector/parallel_vector.cc
```diff
@@ -0,0 +1,270 @@
...
+
+
+std::vector<double> ParallelVector::MakeLocalVector()
+{
+  return std::vector<double>(values_.begin(),
+                             values_.begin() + local_size_);
+}
+
+
+double ParallelVector::operator[](const int64_t local_id) const
```
Naktakala, line 39: This function contradicts its doc string doesn't it?

Naktakala: No it doesn't sorry.

Naktakala (changes requested):

Naktakala (approved):

PR #378, opened 2023-09-22
Nl solver upgrades 2023 09

Naktakala:
## Highlights
- `ParallelVector`:
  - Changed `ParallelVector` to `ParallelSTLVector` 
    (underlying data structures are STL-containers). The STL version using the 
    name `ParallelVector` was using perfect base name. 
  - Defined an abstract base-class `ParallelVector`. Now we can have derived
    classes that differ by their underlying implementation, i.e. `ParallelSTLVector`
    and `ParallelPETScVector`
  - Added more content to the `ParallelSTLVector` and `GhostedParallelSTLVector`
    test to ensure we check the other unused utilities.
  - `VectorGhostCommunicator`: Consolidated all of its members into a single
    data structure that can be initialized and copied more easily.
  - Fixed copy construction operations on ParallelVectors
  - Added norm utility and some basic operations (this will grow to more in the future)
- MPI utilities. Given a communicator we seem to have a lot of code duplicated 
  when getting the current rank off, number of total ranks, and figuring out
  the block offsets for DOFs. Added three utilities for this.
- `chi_math::PETScUtils`. 
  - `CopyVecToSTLvector` by default reallocated the output vector. This screwed
    up cases where it was used for ghosts. Added a boolean flag to control this.
  - `CopyVecToSTLvectorWithGhosts` same.
  - Added `CopySTLvectorToVec`. Don't know why I've always done this manually.
- `chi_math::NonLinearSolver`:
  - Created `NonLinearSolverOptions` block that is based off a `ChiObject`. This
    allows objects using `NonLinearSolver` to more easily obtain and shuttle 
    around options.
  - Changed the constructor of `NonLinearSolver` so that the non-linear method
    comes from the options and not directly through the constructor.
  - Reorganized the declaration of `NonLinearSolver` to have public, protected 
    and private categories in that order.
- `chi_math::SpatialDiscretization`:
  - Added utility method `MakeCellInternalAndBndryNodeIDs`. This was being done
    manually in many places where Dirichlet BCs were involved.
  - `cell_mapping_base.cc` is just pure formatting change.
  - `finite_element` quadrature point data. Added method calls to get the 
    vectors shape-values, shape-grad-values, qpoints-xyz, JxW-values and normals
- `FieldFunction`:
  - `chiExportMultiFieldFunctionToVTK` can now use ff-handles or names.
  - Properly constrained parameter `sdm_type` to the allowed values.
  - `FieldFunctionGridBased::UpdateFieldVector` now just updates local values
    instead of verbatim copying local values (which screwed up when ghosts 
    were used)
- `chi_physics::Solver`
  - The timestepping stuff is still a work in progress.
  - Added a parameter `time_controls` which gets passed to a TimeStepController.
- `chi_physics::TimeStepController`:
  - General enhancements. This is not really used anywhere yet so don't worry 
    too much about it.
- `PostProcessors`:
  - Both `AggregateNodalValue` and `CellVolumeIntegral` now can have a time 
    history which I neglected to put in the first time.
- `ChiObjectFactory`: Added method to get a registered object's input 
  parameters. This makes non-lua factory calls a lot easier.
- `chi_runtime`: Slight improvement to the documentation of the `GetStackItem`
  group of method calls.

## Linear Boltzmann module changes

Adjustments based on the changes to the constructor of `chi_math::NonLinearSolver`


ragusa (approved):

framework/math/NonLinearSolver/NonLinearSolverOptions.cc
```diff
@@ -0,0 +1,98 @@
+#include "NonLinearSolverOptions.h"
+
+#include "ChiObjectFactory.h"
+
+namespace chi_math
+{
+
+RegisterChiObject(chi_math, NonLinearSolverOptions);
```
zachhardy: On a plane, so hard to check, but isn't there a utility for registering _only_ parameters?

framework/math/NonLinearSolver/NonLinearSolverOptions.cc
```diff
@@ -0,0 +1,98 @@
...
+  params.AddOptionalParameter(
+    "nl_abs_tol", 1.0e-8, "Non-linear absolute tolerance");
+  params.AddOptionalParameter(
+    "nl_sol_tol", 1.0e-50, "Non-linear solution tolerance");
+  params.AddOptionalParameter(
+    "nl_max_its", 50, "Non-linear maximum iterations");
+  params.AddOptionalParameter(
+    "nl_max_r_evaluations",
+    -1,
+    "Negative number disables this. The maximum allowed residual evaluations.");
```
zachhardy: Description, then the specific behavior. Switch sentence order.

framework/math/NonLinearSolver/NonLinearSolverOptions.h
```diff
@@ -0,0 +1,39 @@
...
+  double nl_sol_tol_ = 1.0e-50;
+  int    nl_max_its_ = 50;
+  int    nl_max_r_evaluations_ = -1;
+  int    l_max_failed_iterations_ = 1000;
+  double l_rel_tol_ = 1.0e-8;
+  double l_abs_tol_ = 1.0e-8;
+  double l_div_tol_ = 1.0e6;
+  int    l_max_its_ = 100;
+  int    l_gmres_restart_intvl_ = 30;
+  double l_gmres_breakdown_tol_ = 1.0e6;
```
zachhardy, line 34: Would it be better to have a struct for linear and nonlinear options? Alternatively, it would be cool to have a linear solver that the nonlinear solver stores.

framework/math/ParallelVector/ParallelSTLVector.h
```diff
@@ -0,0 +1,158 @@
...
+                    MPI_Comm communicator);
+
+  /**Copy constructor.*/
+  ParallelSTLVector(const ParallelSTLVector& other);
+
+  /**Move constructor.*/
+  ParallelSTLVector(ParallelSTLVector&& other) noexcept;
+
+  std::unique_ptr<ParallelVector> MakeCopy() const override;
+  std::unique_ptr<ParallelVector> MakeNewVector() const override;
```
zachhardy: Could you enlighten me to the purpose of these routines. In particular, why return a unique pointer, and why not just use a copy constructor? What scenario should you use one over the other?  Further, what is the rationale for `MakeNewVector` in addition to `MakeCopy`? To me, the name of the routine isn't suggestive of setting all elements to zero, and even if it was, why can't the user just call `Set` themself?

framework/math/ParallelVector/ParallelVector.h
```diff
@@ -0,0 +1,164 @@
...
+
+  /**Move constructor.*/
+  ParallelVector(ParallelVector&& other) noexcept;
+
+  virtual std::unique_ptr<ParallelVector> MakeCopy() const = 0;
+  virtual std::unique_ptr<ParallelVector> MakeNewVector() const = 0;
+
+  /** Returns the raw stl-vector associated with the local data of this
+   * vector.*/
+  virtual std::vector<double>& RawValues() = 0;
```
zachhardy: Not a big fan of the name. I have always associated `Raw` with pointers to underlying data. Also, what reference STL vector will this return for non-STL parallel vector implementations? What would this return for a `ParallelPETScVector`?

framework/math/ParallelVector/ParallelVector.h
```diff
@@ -0,0 +1,164 @@
...
+   * the local vector.
+   *
+   * \note This accessor allows access to all locally stored elements,
+   *       including any data beyond local_size_ that may exist in derived
+   *       classes.
+   */
+  virtual double& operator[](int64_t local_id) = 0;
+
+  /// Return a vector containing the locally owned data.
+  virtual std::vector<double> MakeLocalVector() = 0;
```
zachhardy, line 80: This is what should be used instead of `RawValues` in its current state (i.e. returning a reference to a STL vector).

framework/math/ParallelVector/ParallelVector.h
```diff
@@ -0,0 +1,164 @@
...
+   */
+  virtual void SetValues(const std::vector<int64_t>& global_ids,
+                         const std::vector<double>& values,
+                         VecOpType op_type) = 0;
+
+  /**In place adding of vectors. The sizes must be compatible.*/
+  virtual void operator+=(const ParallelVector& y) = 0;
+
+  /**Adds a vector multiplied by scalar a. Optimized for a=1.0 and -1.0*/
+  virtual void PlusAY(double a, const ParallelVector& y) = 0;
```
zachhardy:
```suggestion
  virtual void Add(const ParallelVector& y, const double a = 1.0) = 0;
```
What about this with a check for `a = 1.0` to avoid unnecessary multiplication? `PlusAY` doesn't scream add a scaled vector but rather add a matrix vector product. I think avoiding using math notation in function names is better than the alternative. With case rules, it makes it even less clear.

framework/math/ParallelVector/ParallelVector.h
```diff
@@ -0,0 +1,164 @@
...
+
+  /**Sets the local values of one vector equal to another. The sizes must be
+   * compatible.*/
+  virtual void CopyValues(const ParallelVector& y) = 0;
```
zachhardy: This would be better as an overload of `Set` in my opinion. Why introduce a new function name to do the same type of operation in a more particular way?

framework/math/ParallelVector/ParallelVector.h
```diff
@@ -0,0 +1,164 @@
...
+
+  /**Copies a contiguous block of local data from the source vector to the
+   * current vector. The blocks are specified with offset values and num_values.
+   * */
```
zachhardy: I've seen this in deal. The documentation can be improved to say it is copying `num_values` starting from `y_offset` into entries starting at `local_offset`.

framework/math/ParallelVector/ParallelVector.h
```diff
@@ -0,0 +1,164 @@
...
+  virtual void BlockCopyLocalValues(const ParallelVector& y,
+                                    int64_t y_offset,
+                                    int64_t local_offset,
+                                    int64_t num_values) = 0;
+
+  /**Copies a contiguous block of local data from the source STL vector to the
+   * current vector. The local block offset is specified with the local-offset
+   * value. The input STL vector must have num_values entries.
+   * */
```
zachhardy: local_offset, not local-offset. Move */ to end of previous line.

framework/math/ParallelVector/ParallelVector.h
```diff
@@ -0,0 +1,164 @@
...
+  virtual void Assemble() = 0;
+
+  /// Print the local vectors to stings.
+  virtual std::string PrintStr() const = 0;
+
+  /**
+   * Communicate the current ghost entries, if applicable, to all other
+   * processes to update the locally stored ghost data.
+   */
+  virtual void CommunicateGhostEntries(){};
```
zachhardy, line 168:
```suggestion
  /**
   * Communicate the current ghost entries, if applicable, to all other
   * processes to update the locally stored ghost data.
   */
  virtual void CommunicateGhostEntries() {};
```
We've come full circle from drawing a line in the sand with using the word "ghost" in a non-ghost vector class

framework/math/ParallelVector/ParallelVector.h
```diff
@@ -0,0 +1,164 @@
...
+  virtual void PlusAY(double a, const ParallelVector& y) = 0;
+
+  /**Sets the local values of one vector equal to another. The sizes must be
+   * compatible.*/
+  virtual void CopyValues(const ParallelVector& y) = 0;
+
+  /**Copies a contiguous block of local data from the source vector to the
+   * current vector. The blocks are specified with offset values and num_values.
+   * */
+  virtual void BlockCopyLocalValues(const ParallelVector& y,
```
zachhardy: `Set` overload.

framework/math/ParallelVector/ParallelVector.h
```diff
@@ -0,0 +1,164 @@
...
+                                    int64_t y_offset,
+                                    int64_t local_offset,
+                                    int64_t num_values) = 0;
+
+  /**Copies a contiguous block of local data from the source STL vector to the
+   * current vector. The local block offset is specified with the local-offset
+   * value. The input STL vector must have num_values entries.
+   * */
+  virtual void BlockCopyLocalValues(const std::vector<double>& y,
```
zachhardy: `Set` overload.

framework/math/SpatialDiscretization/CellMappings/cell_mapping_base.cc
```diff
@@ -6,12 +6,13 @@
...
-  CellMapping(const chi_mesh::MeshContinuum& in_grid,
-              const chi_mesh::Cell& in_cell,
-              size_t in_num_nodes,
-              std::vector<std::vector<int>> in_face_node_mappings,
-              const VandAFunction& volume_area_function) : ref_grid_(in_grid),
+chi_math::CellMapping::CellMapping(
+  const chi_mesh::MeshContinuum& in_grid,
+  const chi_mesh::Cell& in_cell,
+  size_t in_num_nodes,
+  std::vector<std::vector<int>> in_face_node_mappings,
```
zachhardy, line 13: We should move away from these `in_` arguments as we encounter them. These are not necessary now that members use a trailing underscore.

framework/math/VectorGhostCommunicator/vector_ghost_communicator.h
```diff
@@ -28,35 +34,40 @@ class VectorGhostCommunicator
...
-  std::vector<int> senddispls_;
-  std::vector<int> recvcounts_;
-  std::vector<int> recvdispls_;
+    std::vector<int64_t> local_ids_to_send_;
+    std::map<int64_t, size_t> ghost_to_recv_map_;
+  };
 
-  std::vector<int64_t> local_ids_to_send_;
-  std::map<int64_t, size_t> ghost_to_recv_map_;
+  const CachedParallelData cached_parallel_data_;
```
zachhardy, line 66: Consider declaring the struct at the top of the class.

framework/math/VectorGhostCommunicator/vector_ghost_communicator.h
```diff
@@ -28,35 +34,40 @@ class VectorGhostCommunicator
...
 
+private:
+  int FindOwnerPID(int64_t global_id) const;
+  CachedParallelData MakeCachedParallelData();
```
zachhardy, line 70: Consider moving up to be with the other routines.

framework/mpi/chi_mpi_utils.h
zachhardy, line 1: Consider adding the all to all stuff into these files.

framework/physics/FieldFunction/ff_gridbased_00_constrdestr.cc
```diff
@@ -24,6 +24,10 @@ chi::InputParameters FieldFunctionGridBased::GetInputParameters()
   params.AddOptionalParameter(
     "initial_value", 0.0, "The initial value to assign to the field function");
 
+  using namespace chi_data_types;
+  params.ConstrainParameterRange(
+    "sdm_type", AllowableRangeList::New({"FiniteVolume", "PWLC", "PWLD"}));
```
zachhardy: Should we be consistent and use "FV" for finite volume?

framework/physics/SolverBase/chi_solver.cc
```diff
@@ -32,6 +32,11 @@ chi::InputParameters Solver::GetInputParameters()
   params.AddOptionalParameter(
     "timestep_controller", 0, "Timestep controller to use for timestepping.");
 
+  params.AddOptionalParameterBlock(
+    "time_controls",
+    chi::ParameterBlock{},
+    "Parameters to pass to the time_controller.");
```
zachhardy: This confuses me. What are `time_controls` and a `time_controller`? Should this be `timestep_controls`? If so, what is the idea? A user can pass I assume a handle to a `timestep_controller` or parameters for one? I would just require a handle to a controller instead of this. Please add that the `timestep_controller` input is a handle to a controller.

framework/physics/SolverBase/chi_solver.cc
```diff
@@ -48,10 +53,13 @@ Solver::Solver(const chi::InputParameters& params)
...
 
-  if (not user_params.Has("timestep_controller"))
+  const bool has_controller = user_params.Has("timestep_controller");
+  const bool has_time_controls = user_params.Has("time_controls");
+
+  if (not has_controller)
   {
     chi::ParameterBlock ts_params;
-    ts_params.AddParameter("initial_dt", dt_);
+    if (has_time_controls) ts_params = user_params.GetParam("time_controls");
```
zachhardy: Same as above

framework/physics/SolverBase/chi_solver.cc
```diff
@@ -64,6 +72,10 @@ Solver::Solver(const chi::InputParameters& params)
     time_step_controller_ = Chi::GetStackItemPtrAsType<TimeStepController>(
       Chi::object_stack, handle, __FUNCTION__);
   }
+
+  ChiInvalidArgumentIf(has_controller and has_time_controls,
+                       "Both \"has_controller\" and \"has_time_controls\" "
+                       "cannot be specified at the same time.");
```
zachhardy: Unnecessary if you just require a controller handle.

framework/physics/SolverBase/chi_solver.cc
```diff
@@ -48,10 +53,13 @@ Solver::Solver(const chi::InputParameters& params)
 {
   const auto& user_params = params.ParametersAtAssignment();
 
-  if (not user_params.Has("timestep_controller"))
+  const bool has_controller = user_params.Has("timestep_controller");
+  const bool has_time_controls = user_params.Has("time_controls");
```
zachhardy: Same as above

framework/physics/TimeStepControllers/TimeStepController.cc
```diff
@@ -1,28 +1,115 @@
...
+  params.AddOptionalParameter("end_time", 1.0, "End time");
+  params.AddOptionalParameter(
+    "max_time_steps",
+    -1,
+    "Maximum number of timesteps to take. A negative number disables this.");
+
+  params.AddOptionalParameter("eps",
+                              1.0e-12,
+                              "General time tolerance. This is used to "
+                              "constrict the dt to the end-time.");
```
zachhardy:
```suggestion
                              "constrict the dt to the final time.");
```
Either way, no -

framework/physics/TimeStepControllers/TimeStepController.cc
```diff
@@ -1,28 +1,115 @@
...
+}
+
+/**Determines if the time is at the end time (within tolerance) or greater.*/
+bool TimeStepController::AtEndTime() const
+{
+  if (std::fabs(time_ - end_time_) < general_tolerance_)
+    return true;
+
+  if (time_ > end_time_)
+    return true;
```
zachhardy: This seems incompatible with the routine name and docstring, what is its purpose? Is ending at `end_time_` optional?

framework/physics/TimeStepControllers/TimeStepController.cc
```diff
@@ -1,28 +1,115 @@
...
+
+/**Provides the equivalent timestep for ending at the end-time.*/
+double TimeStepController::StepSizeLimitedToEndTime() const
 {
+  double dt_limited = dt_;
+
+  if ((end_time_ - time_) <= dt_) dt_limited = end_time_ - time_;
+
+  return dt_limited;
+}
```
zachhardy: What is the purpose for this?

framework/physics/TimeStepControllers/TimeStepController.cc
```diff
@@ -1,28 +1,115 @@
...
+    end_time_(params.GetParamValue<double>("end_time")),
+    max_time_steps_(params.GetParamValue<int>("max_time_steps")),
+    general_tolerance_(params.GetParamValue<double>("eps"))
+{
+}
+
+/**Overridable method to get the timestep size.*/
+double TimeStepController::GetTimeStepSize()
+{
+  dt_ = StepSizeLimitedToEndTime();
```
zachhardy: An accessor changing a member??? Not a fan.

framework/physics/TimeStepControllers/TimeStepController.h
```diff
@@ -6,18 +6,59 @@
...
+
+  /**Builds a formatted string of the time information.*/
+  std::string StringTimeInfo() const;
+
 protected:
   static chi::InputParameters GetInputParameters();
   explicit TimeStepController(const chi::InputParameters& params);
 
-  double current_timestep_size_;
+  /**Provides the equivalent timestep for ending at the end-time.*/
```
zachhardy:
```suggestion
  /**Provides the equivalent timestep for ending at the final time.*/
```

framework/post_processors/CellVolumeIntegralPostProcessor.cc
```diff
@@ -145,6 +145,21 @@ void CellVolumeIntegralPostProcessor::Execute(const Event& event_context)
...
     value_ = ParameterBlock("", globl_integral / globl_volume);
   }
+
+  const int event_code = event_context.Code();
+  if (event_code == 32 /*SolverInitialized*/ or
+      event_code == 38 /*SolverAdvanced*/)
+  {
+    const auto& event_params = event_context.Parameters();
+
+    if (event_params.Has("timestep_index") and event_params.Has("timestamp"))
```
zachhardy: Is `timestamp` the same as `time`? If so, I would recommend changing `timestep_index` to `timestamp_index` to be more clear.

framework/post_processors/CellVolumeIntegralPostProcessor.cc
```diff
@@ -145,6 +145,21 @@ void CellVolumeIntegralPostProcessor::Execute(const Event& event_context)
...
+    {
+      const size_t index = event_params.GetParamValue<size_t>("timestep_index");
```
zachhardy, line 157: Same as above

framework/math/NonLinearSolver/non_linear_solver.h
zachhardy, line 1: Should this be changed to `NonLinearSolver.h`?

framework/math/NonLinearSolver/non_linear_solver_petsc.cc
zachhardy, line 1: Should this be changed to `NonLinearSolverPETSc.cc`?

zachhardy (changes requested):
See comments. Lots of changes in different areas in this PR so please forgive any comments that don't seem to understand the big picture.

I noticed that one quite a few files you did a reformat according to the `.clang_format` file while on others you didn't. We should standardize the whole code base with whatever we decide the appropriate formatting options are and then require future PRs to conform to them.

framework/math/NonLinearSolver/NonLinearSolverOptions.cc
Naktakala: There is. When you just have a lonesome `GetInputParameters` call. In this case there is an actual constructor that takes this value,
```
explicit NonLinearSolver(NLSolverContextPtr context_ptr,
                           const chi::InputParameters& params =
                             NonLinearSolverOptions::GetInputParameters())
    : solver_name_(params.GetParamValue<std::string>("name")),
      context_ptr_(context_ptr),
      options_(params)
  {
  }
```

However, looking at it, I don't see an immediate need to be able to create the object standalone. I think we can just register the syntax.

Naktakala: Changed to `RegisterChiObjectParametersOnly`

framework/math/NonLinearSolver/NonLinearSolverOptions.cc
Naktakala: Done

framework/math/NonLinearSolver/NonLinearSolverOptions.h
Naktakala: I have that of something like this. It becomes tricky when you want it to appear completely pluggable because PETSc is so different from other design patterns.

We can easily define LinearSolverOptions and then make NonLinearSolverOptions inherit from it. However, that would be a scope for a different PR.

framework/math/NonLinearSolver/non_linear_solver_petsc.cc
Naktakala: Yes. Renamed.

framework/math/NonLinearSolver/non_linear_solver.h
Naktakala: Yes. Renamed.

framework/math/ParallelVector/ParallelSTLVector.h
Naktakala: Oh damn. Base class missing documentation. Apologies.

```c++
/**Creates a copy of the vector datastructures AND values.
  * This routine requires no communication.*/
  virtual std::unique_ptr<ParallelVector> MakeCopy() const = 0;

  /**Creates a copy of the vector datastructures but NOT the values. The
  * values are defaulted to zero. This routine requires no communication.*/
  virtual std::unique_ptr<ParallelVector> MakeNewVector() const = 0;
```

Naktakala: This is akin to PETSc's `VecCopy` and `VecClone` and they are very useful. Consider the code below:
```
r_t_tp1_ = r.MakeNewVector();
SetModeToTimeOnly();
eq_system_->ComputeResidual(x, *r_t_tp1_);

if (TimeIDListHasID(r_time_ids, TimeID::T_PLUS_1))
{
    r_x_tp1_ = r.MakeNewVector();
    SetModeToNonTimeOnly();
    eq_system_->ComputeResidual(x, *r_x_tp1_);
    r_map_[TimeID::T_PLUS_1] = (&(*r_x_tp1_));
}
```

Naktakala: This is what a MakeCopy, like you suggest will look like:
```
r_t_tp1_ = r.MakeCopy();
r_t_tp1.Set(0.0);

SetModeToTimeOnly();
eq_system_->ComputeResidual(x, *r_t_tp1_);

if (TimeIDListHasID(r_time_ids, TimeID::T_PLUS_1))
{
    r_x_tp1_ = r.MakeNewVector();
    r_x_tp1_.Set(0.0);
    SetModeToNonTimeOnly();
    eq_system_->ComputeResidual(x, *r_x_tp1_);
    r_map_[TimeID::T_PLUS_1] = (&(*r_x_tp1_));
}
```

Naktakala: Another example:
```
std::vector<std::unique_ptr<ParallelVector>>
EquationSystem::InitSolutionHistory()
{
  std::vector<std::unique_ptr<ParallelVector>> old_solution_vectors;
  for (size_t t = 0; t < num_solution_histories_; ++t)
  {
    auto vec = main_solution_vector_->MakeNewVector();
    old_solution_vectors.push_back(std::move(vec));
  }

  return old_solution_vectors;
}
```

framework/math/ParallelVector/ParallelVector.h
Naktakala: Please suggest an alternative name. With ParallelPETScVector there would have to be the notion of open/closed like other codes do. For now the code on the base class just throws "Method not implemented".

Let me think about a possible alternative here.

framework/math/ParallelVector/ParallelVector.h
Naktakala: Well, this goes with an alternative that I am thinking about. The make local vector business for setting the local values of a vector requires 2 copies. Once to copy out of the vector, then you use something like CopyVecToSTLVec, then you have to use Set again which does bounds checking AND another copy.

The real solution here would be to overload the `CopyValues` routine to operate on a PETSc vector.

framework/math/ParallelVector/ParallelVector.h
Naktakala: Ok. We can do this. I want to eventually get to all of these:
image

framework/math/ParallelVector/ParallelVector.h
Naktakala: I've already messaged you in private and said I conceded. If you want a public concession then here it is. Congrats Zach, our disagreement tilted in your favor with me self-correcting that it is a better idea. Now let's do better next time and not put things like this in a review.

framework/math/SpatialDiscretization/CellMappings/cell_mapping_base.cc
Naktakala: Looks like I accidentally hit the reformat short-cut on this file. I might have added something then removed it.

I agree with your comment, and take responsibility for it, however, this can send this PR down a rabbit hole and changed into a "fix the in_ business". I would love help in committing a single PR with just formatting changes.

framework/math/VectorGhostCommunicator/vector_ghost_communicator.h
Naktakala: Please clarify. The file is currently structured public, protected, private. The struct should maintain it's protected status. What exactly is needed?

framework/math/VectorGhostCommunicator/vector_ghost_communicator.h
Naktakala: Please clarify as well. 
Option A : Methods public+protected+private --- Members public+protected+private
Option B : public Method+Members --- protected Method+Members --- private Method+Members

?

framework/mpi/chi_mpi_utils.h
Naktakala: Can't. Its a fully defined templated class.

framework/physics/FieldFunction/ff_gridbased_00_constrdestr.cc
Naktakala: I've spent some thought on this. Consider this: I thought about likely new additions to the spatial discretization family and Lagrange came to mind as the likely next candidate. In fact we could have a hybrid Lagrange but PWL sprinkled in if a polygon or polyhedron is in there. With that in regard I singled out the more wordy SDMs and shortened them. So Instead of having `FV, PWLC, PWLD, LG?` we would have `FiniteVolume, Lagrange, PWLC, PWLD, DiscontinuousLagrange, Monomial, Raviart-Thomas, etc.` 

Would be a shame to have `PieceWiseLinear` and `PieceWiseLinearDiscontinuous` in there for the sake of consistency, however, I could live with it. Would you like me then to write them all out?

framework/physics/SolverBase/chi_solver.cc
Naktakala: The paradigm I am aiming for here is to have the TimeStepController be in control of time in its entirety. Including termination times, whether (among solvers) time should synchronize, etc. 

(note: we could call it TimeController but that clashes with the succinct nature of things like AdaptiveTimeStepController)

The user then has two options, add the necessary options for a default TimeStepController, or pass a new TimeStepController, both accept time_controls as arguments, albeit, some controller might have more extensive control options than others.

Adjusted the documentation to:
```c++
params.AddOptionalParameter(
    "timestep_controller",
    0,
    "Timestep controller to use for timestepping. "
    "Cannot be supplied if the time_controls parameter is supplied");

  params.AddOptionalParameterBlock(
    "time_controls",
    chi::ParameterBlock{},
    "Parameters to pass to the default time_controller. If a "
    "timestep_controller is specified then this parameter can not be "
    "specified.");
  params.LinkParameterToBlock("time_controls",
                              "chi_physics::TimeStepController");
```

framework/physics/SolverBase/chi_solver.cc
Naktakala: But thats the whole point. We don't just require a handle. We CAN accept just the basic controls.

framework/physics/SolverBase/chi_solver.cc
Naktakala: Response same as above.

framework/physics/SolverBase/chi_solver.cc
Naktakala: Response same as above.

framework/math/ParallelVector/ParallelVector.h
Naktakala:
```
/**Copies a contiguous block of local data (num_values entries) from the
  * source vector (starting at y_offset) to the
   * current vector starting at local_offset. */
  virtual void BlockCopyLocalValues(const ParallelVector& y,
                                    int64_t y_offset,
                                    int64_t local_offset,
                                    int64_t num_values) = 0;

  /**Copies a contiguous block of data from the source STL vector to the
   * current vector starting at local_offset. The input STL vector must have 
   * exactly num_values entries.
   * */
  virtual void BlockCopyLocalValues(const std::vector<double>& y,
                                    int64_t local_offset,
                                    int64_t num_values) = 0;
```

framework/math/ParallelVector/ParallelVector.h
Naktakala: Done

framework/physics/FieldFunction/ff_gridbased_00_constrdestr.cc
Naktakala: It's safe for me to say that I don't think the all-abbreviated case would be good.
Which option do you prefer here:
```lua
tbulk = chi_physics.FieldFunctionGridBased.Create({
  name = "T_bulk",
  --sdm_type = "FiniteVolume",
  --sdm_type = "PiecewiseLinear",
  --sdm_type = "PiecewiseLinearDiscontinuous",
  --sdm_type = "Lagrange",
  --sdm_type = "LagrangeDiscontinuous"
  --sdm_type = "Monomial"
  --sdm_type = "RaviartThomas"
  initial_value = 50.0
})

tbulk = chi_physics.FieldFunctionGridBased.Create({
  name = "T_bulk",
  --sdm_type = "FiniteVolume",
  --sdm_type = "PWLC",
  --sdm_type = "PWLD",
  --sdm_type = "Lagrange",
  --sdm_type = "LagrangeD"
  --sdm_type = "Monomial"
  --sdm_type = "RaviartThomas"
  initial_value = 50.0
})
```

framework/physics/TimeStepControllers/TimeStepController.cc
Naktakala: I actually agree on this. The next timestep size should be determined during the Advance. Given that nothing right now uses this and that it is still a work in progress, would you be fine with deferring this till future work?With better examples.

framework/physics/TimeStepControllers/TimeStepController.cc
Naktakala: Along the lines of future work. There might be a need for a distinct `end_time` and `final_time`. The `end_time` gets used in a multiphysics setting whilst the `final_time` is the true final.

A multiphysics controller can then allow something like thermal-hydraulics to sub-step within the current time step until it reaches the `end-time`. And all of this needs to also support adaptivity. All of the solvers need to completely end at `final_time`.

I don't have the details fleshed out here yet/

framework/physics/TimeStepControllers/TimeStepController.cc
Naktakala: I am still working on this. Technically other logic should prevent overstepping the end-time.  This is just a temporary fail-safe.

framework/physics/TimeStepControllers/TimeStepController.h
Naktakala: Difference end_time versus final_time but still work in progress.

framework/post_processors/CellVolumeIntegralPostProcessor.cc
Naktakala: Changed `timestamp` to `time`

framework/post_processors/CellVolumeIntegralPostProcessor.cc
Naktakala: Response same as above

framework/physics/FieldFunction/ff_gridbased_00_constrdestr.cc
Naktakala: Regardless, the user would have to consult documentation. I am just rooting for not spelling out Discontinuououuoss every time.

framework/math/ParallelVector/ParallelVector.h
Naktakala: We should have a discussion on this

framework/math/ParallelVector/ParallelVector.h
Naktakala: The overall theme here, and the comments below, seem to be to favor overloads. We should discuss this but my opinion is that it is much harder to search through the documentation of overloads than it is to look for hints of the intent through the method name. You basically get additional short-hand information folded into the name of the method.

In this case I don't think `Add` here is a necessarily a bad option. I just think that adding more functionality where the statement reads like an equation is better. Examples:
```
x.Add(y, 2.0) --> x add y 2.0
x.PlusAy(2.0, y) --> x plus ay, a=2.0

x.Add({y1,y2,y3}, {1.0,2.0,3.0})
x.PlusMultiAY({1.0,2.0,3.0}, {y1,y2,y3})
```
as long as we are consistent the A one can eventually get a feel for the method name without looking it up (Like for PETSc).

framework/math/ParallelVector/ParallelVector.h
Naktakala: IDK `x.Set(y)` has a literal feeling to it, for me. Set it to the same vector? Set it to the same value? What if x is a non-ghosted vector and y is ghosted, what will x be afterwards?

framework/math/ParallelVector/ParallelVector.h
Naktakala: Losing a lot of short-hand info. If we have that many overloaded `Set`, you can basically get very little from auto-complete.

framework/math/ParallelVector/ParallelVector.h
Naktakala: Response same as above.

framework/physics/TimeStepControllers/TimeStepController.cc
Naktakala: Redocumented to:
```
params.AddOptionalParameter("eps",
                              1.0e-12,
                              "General time tolerance. This is used in "
                              "fuzzy equals.");
```

framework/math/NonLinearSolver/NonLinearSolverOptions.h
zachhardy: This may merit more discussion offline as that I'm not sure what you really mean by "completely pluggable".

framework/math/ParallelVector/ParallelSTLVector.h
zachhardy: I'm still not sure the purpose of returning pointers instead of objects here.

framework/math/ParallelVector/ParallelVector.h
Naktakala: image

zachhardy: Okay, so I am sensing that we want to reimplement most/all the PETSc functions with effectively the same names. If that is the case, put the actual equations being carried out in the documentation as PETSc does. This helps a user who isn't used to `AY` being a scalar times a vector. Also, if someone is trying to add vectors, I would imaging a user looks up `Add` over `Plus`, but that's just me. I'm fine with whatever convention you choose here, but I would add equations to documentation.

framework/math/ParallelVector/ParallelVector.h
zachhardy: I'm not a fan of having a gazillion different function names to do extremely similar things. In my opinion, it makes it harder in the documentation not easier. If someone wants to set a vector to something, it is useful to have all of the different ways a vector can be set next to one another. Does `x.Set(1.0)` feel like the vector becomes a scalar value to you? Why not have `x.SetAllValuesToScalar(1.0)` and `x.SetToSTLVector(y)`, and `x.SetToOtherParallelVector(y)`?

framework/math/ParallelVector/ParallelVector.h
zachhardy: I think we should cross the PETSc wrapper when we get to it instead of trying to build it all in before it exists. I'd rather see changes for compatibility than to try to imagine what a PETSc implementation would look like and make a judgement based on that.

framework/math/PETScUtils/petsc_utils_04_vecmanip.cc
```diff
@@ -2,54 +2,72 @@
...
-void chi_math::PETScUtils::CopyVecToSTLvector(
-  Vec x, std::vector<double>& data, size_t N)
+void chi_math::PETScUtils::CopyVecToSTLvector(Vec x,
+                                              std::vector<double>& data,
+                                              size_t N,
+                                              bool resize_STL /*=true*/)
 {
-  data.clear();
-  data.resize(N,0.0);
+  if (resize_STL)
```
wdhawkins, line 13: If resize_STL is false, we should either assert(data.size() == N) or throw an error if data.size() != N.

framework/math/ParallelVector/ParallelVector.h
zachhardy: Once you have `x.Set(` all of the different overloads show up in CLion and I'm sure in VSCode. This is something we've all seen when calling things like `x.resize(` on an STL vector in your favorite IDE.

framework/math/ParallelVector/ParallelVector.h
zachhardy: Me too

framework/math/PETScUtils/petsc_utils_04_vecmanip.cc
```diff
@@ -2,54 +2,72 @@
...
+    data.clear();
+    data.assign(N, 0.0);
+  }
+
   const double* x_ref;
-  VecGetArrayRead(x,&x_ref);
+  VecGetArrayRead(x, &x_ref);
 
-  for (size_t i=0; i<N; ++i)
+  for (size_t i = 0; i < N; ++i)
```
wdhawkins: std::copy might be better here.

I would do something like:

```
if (resize_STL)
{
   data.clear();
   data.assign(N, 0.0);
}
assert(data.size() == N)

const double* x_ref;
VecGetArrayRead(x, &x_ref);
std::copy_n(x_ref, N, data.data());
VecRestoreArray(x, &x_ref);`
```

A couple of side notes:
1) Might be worth templating on type.
2) I don't know anything about the PETSC calls.  It might be worth setting x_ref to nullptr and checking it before the copy depending on what PETSC does if the call to VecGetArray fails.

Note all that code above is untested... I think it'll work.

framework/math/VectorGhostCommunicator/vector_ghost_communicator.h
zachhardy: Nothing is needed, just formatting suggestion. I always like to do 
```
class <class_name>
{
public/protected/private:
  typedefs ...
public/protected/private:
  structs ...
public/protected/private:
  functions ...
public/protected/private:
  members ...
};
```
instead of organizing by public/protected/private.

framework/math/PETScUtils/petsc_utils_04_vecmanip.cc
```diff
@@ -2,54 +2,72 @@
...
+// ###################################################################
 /**Copies a PETSc vector to a STL vector. Only the local portion is
  * copied.*/
 void chi_math::PETScUtils::CopyVecToSTLvectorWithGhosts(
-  Vec x, std::vector<double>& data, size_t N)
+  Vec x, std::vector<double>& data, size_t N, bool resize_STL /*=true*/)
 {
-  data.clear();
-  data.resize(N,0.0);
+  if (resize_STL)
```
wdhawkins, line 37: Same comment as above.  Needs assert() or throw an error if data.size() != N.

framework/math/VectorGhostCommunicator/vector_ghost_communicator.h
zachhardy: See above

framework/physics/FieldFunction/ff_gridbased_00_constrdestr.cc
zachhardy: My thought is that we should abbreviate for now, and then revisit when we have those other options available. I would lean towards "standard" methods being abbreviated (i.e. methods often referred to with an abbreviation) and non-standard methods being spelt out.

framework/physics/SolverBase/chi_solver.cc
zachhardy: I understand the idea, but I don't understand your "succinct nature of things" comment. I think if the `timestep_controller` controls time and doesn't just spit out a time step then it should be called `time_controller`. 

I still think that this is a weird input language and that the user should just be required to pass a `time_controller` handle to the solver, default or not. The user has to create the mesh outside of the solver, why not the time stepper too? It makes for a cleaner input in my opinion.

framework/math/PETScUtils/petsc_utils_04_vecmanip.cc
Naktakala: Could you give some protocode here? No idea how to do this with that pointer-based vector

framework/physics/TimeStepControllers/TimeStepController.cc
zachhardy: Uh, okay?

framework/math/PETScUtils/petsc_utils_04_vecmanip.cc
Naktakala: Good recommendation. I would ammend it to data.size() >= N because you could pass a part of ghosted vector.

framework/math/PETScUtils/petsc_utils_04_vecmanip.cc
Naktakala: Got it. This one would be .size() == N

framework/physics/TimeStepControllers/TimeStepController.cc
zachhardy: I guess. I don't think we should introduce features that are both unused and in a state that needs changing. PRs should not create issues, but rather resolve them.

framework/math/PETScUtils/petsc_utils_04_vecmanip.cc
Naktakala: How does one do that with the double* business?

zachhardy: Maybe try `std::copy(x_ref, x_ref + N, data.begin());`?

framework/math/ParallelVector/ParallelVector.h
Naktakala: I am sorry I just don't agree. BlockSet here is a way better option and pops out with the autocomplete as "Yes, thats what I want".

framework/physics/FieldFunction/ff_gridbased_00_constrdestr.cc
Naktakala: Ok. I will then make it all abbreviated. Just bare in mind that Lagrange is pretty close to being reality so then I would prefer `Langrange` and `LagrangeD`

framework/math/ParallelVector/ParallelVector.h
zachhardy: Okay, that's fine

framework/math/ParallelVector/ParallelVector.h
zachhardy: I would change the name of this to `CopyLocalValues`.

framework/math/ParallelVector/ParallelSTLVector.h
Naktakala: From the Slack discussion:
The problem I encountered is that when are in a routine where you passed the base ParallelVector by reference, you can't copy it. Its an abstract class. There is no copy constructor defined and you cannot make a virtual one because ParallelSTL vector or GhostedParallelSTL vector doesn't know what its dealing with.

Petsc hides this very well from us because they're Vec  is actually a typedef on a pointer. Inherently we should always store our vectors as pointers for polymorphism

Now, why a unique_ptr. 1) The ownership is very clear. The vector that spawns a new vector should no longer hold on to it so it shouldn't be a smart_pointer. 2) A unique_ptr can easily be converted to a shared_ptr but a shared_ptr cannot be converted to a shared_ptr.

framework/math/ParallelVector/ParallelVector.h
Naktakala: Reworked to this:
```
/** Returns a direct pointer to the memory array used internally by the vector
   * to store its owned elements*/
  virtual double* Data() = 0;

  /** Returns a direct const pointer to the memory array used internally by the
  * vector to store its owned elements*/
  virtual const double* Data() const = 0;
```

framework/math/ParallelVector/ParallelVector.h
Naktakala: We now have `double* ParallelVector::Data()` which serves all our needs. It's pretty efficient too.

framework/math/ParallelVector/ParallelVector.h
Naktakala: Done

framework/physics/FieldFunction/ff_gridbased_00_constrdestr.cc
Naktakala: Done

framework/physics/TimeStepControllers/TimeStepController.cc
Naktakala: Fixed. Have a look at the new implementation and example usage in the point-reactor module.

framework/math/PETScUtils/petsc_utils_04_vecmanip.cc
Naktakala: Done

framework/math/PETScUtils/petsc_utils_04_vecmanip.cc
```diff
@@ -2,54 +2,78 @@
...
 
-//###################################################################
+void chi_math::PETScUtils::CopySTLvectorToVec(const std::vector<double>& data,
+                                              Vec x,
+                                              size_t N)
+{
+  double* x_ref;
+  VecGetArray(x, &x_ref);
+
+  for (size_t i = 0; i < N; ++i)
```
wdhawkins: Same here as above with std::copy.

framework/math/PETScUtils/petsc_utils_04_vecmanip.cc
```diff
@@ -58,31 +82,31 @@ void chi_math::PETScUtils::CopyGlobalVecToSTLvector(
...
   //=================================== Copy to STL
   data.clear();
-  data.resize(N,0.0);
+  data.resize(N, 0.0);
   const double* x_ref;
-  VecGetArrayRead(local_vec,&x_ref);
+  VecGetArrayRead(local_vec, &x_ref);
 
-  for (size_t i=0; i<N; ++i)
+  for (size_t i = 0; i < N; ++i)
```
wdhawkins: Another copy. Definitely a fan of stl where we can use it.

framework/physics/TimeStepControllers/ConstantTimeStepper.cc
```diff
@@ -0,0 +1,27 @@
...
+
+RegisterChiObject(chi_physics, ConstantTimeStepper);
+
+chi::InputParameters ConstantTimeStepper::GetInputParameters()
+{
+  chi::InputParameters params = TimeStepper::GetInputParameters();
+
+  params.SetGeneralDescription(
+    "Timestep controller that does not dynamically change.");
+  params.SetDocGroup("doc_TimeStepControllers");
```
zachhardy: Should we rename this to `doc_TimeSteppers`?

framework/math/PETScUtils/petsc_utils_04_vecmanip.cc
```diff
@@ -36,6 +40,10 @@ void chi_math::PETScUtils::CopyVecToSTLvectorWithGhosts(
     data.clear();
     data.assign(N, 0.0);
   }
+  else
+    ChiLogicalErrorIf(data.size() != N,
+                      "data.size() != N, " + std::to_string(data.size()) +
+                        " < " + std::to_string(N));
```
zachhardy, line 45: This doesn't technically need to be within the `else` but it doesn't change functionality with it there.

framework/math/ParallelVector/ParallelSTLVector.cc
```diff
@@ -79,12 +82,84 @@ double& ParallelSTLVector::operator[](const int64_t local_id)
...
 {
-  ChiInvalidArgumentIf(local_vector.size() != local_size_,
+  ChiInvalidArgumentIf(local_vector.size() < local_size_,
                        "Incompatible local vector size.");
 
-  values_ = local_vector;
+  // We cannot assign values = local_vector because this might
+  // destroy the internals of a ghosted vector
+  for (size_t i = 0; i < local_size_; ++i)
+    values_[i] = local_vector[i];
```
zachhardy: Since we are doing it elsewhere we could use `std::copy` here... but I will leave this up to you. I am fine either way. Just depends on whether we want to bulk do that everywhere we can or not.

framework/math/ParallelVector/ParallelSTLVector.cc
```diff
@@ -79,12 +82,84 @@ double& ParallelSTLVector::operator[](const int64_t local_id)
...
+
+  const int64_t local_end = local_offset + num_values;
+  ChiInvalidArgumentIf(
+    local_end > local_size_,
+    "local_offset + num_values=" + std::to_string(local_end) +
+      ", is out of range for destination vector with local size " +
+      std::to_string(local_size_));
+
+  for (int64_t i = 0; i < num_values; ++i)
+    values_[local_offset + i] = y[i];
```
zachhardy: See above about `std::copy`.

framework/math/ParallelVector/ParallelSTLVector.cc
```diff
@@ -79,12 +82,84 @@ double& ParallelSTLVector::operator[](const int64_t local_id)
...
+                    "y.LocalSize() != local_size_");
+  ChiLogicalErrorIf(y.GlobalSize() != global_size_,
+                    "y.GlobalSize() != global_size_");
+
+  // This prevents us from unnecessarily going through the operator[]
+  // of y, which again applies bounds checking
+  const double* y_data = y.Data();
+
+  for (int64_t i = 0; i < local_size_; ++i)
+    values_[i] = y_data[i];
```
zachhardy: See above about `std::copy`.

framework/math/ParallelVector/ParallelSTLVector.cc
```diff
@@ -79,12 +82,84 @@ double& ParallelSTLVector::operator[](const int64_t local_id)
...
+    "local_offset + num_values=" + std::to_string(local_end) +
+      ", is out of range for destination vector with local size " +
+      std::to_string(local_size_));
+
+  // This prevents us from unnecessarily going through the operator[]
+  // of y, which again applies bounds checking
+  const double* y_data = y.Data();
+
+  for (int64_t i = 0; i < num_values; ++i)
+    values_[local_offset + i] = y_data[y_offset + i];
```
zachhardy: See above about `std::copy`.

framework/math/ParallelVector/ParallelSTLVector.cc
```diff
@@ -124,104 +199,61 @@ void ParallelSTLVector::operator+=(const ParallelVector& y)
...
-  ChiLogicalErrorIf(y.GlobalSize() != global_size_,
-                    "y.GlobalSize() != global_size_");
+  // This prevents us from unnecessarily going through the operator[]
+  // of y, which again applies bounds checking
+  const double* y_data = y.Data();
 
-  for (size_t i = 0; i < local_size_; ++i)
-    values_[i] += y.values_[i];
+  for (int64_t i = 0; i < local_size_; ++i)
+    values_[i] += y_data[i];
```
zachhardy, line 207: If you want to be fancy, an `std::transform` would work here with a lambda.

framework/math/ParallelVector/ParallelSTLVector.cc
```diff
@@ -124,104 +199,61 @@ void ParallelSTLVector::operator+=(const ParallelVector& y)
...
   else if (a == -1.0)
-    for (size_t i = 0; i < local_size_; ++i)
-      values_[i] -= y[scint64_t(i)];
+    for (int64_t i = 0; i < local_size_; ++i)
+      values_[i] -= y_data[i];
   else
-    for (size_t i = 0; i < local_size_; ++i)
-      values_[i] += a * y[scint64_t(i)];
+    for (int64_t i = 0; i < local_size_; ++i)
+      values_[i] += a * y_data[i];
```
zachhardy: See comment above about `std::transform`.

framework/math/ParallelVector/ParallelSTLVector.cc
```diff
@@ -124,104 +199,61 @@ void ParallelSTLVector::operator+=(const ParallelVector& y)
...
-{
-  ChiLogicalErrorIf(y.LocalSize() != local_size_,
-                    "y.LocalSize() != local_size_");
-  ChiLogicalErrorIf(y.GlobalSize() != global_size_,
-                    "y.GlobalSize() != global_size_");
-
-  for (size_t i = 0; i < local_size_; ++i)
-    values_[i] = y.values_[i];
+  for (int64_t i = 0; i < local_size_; ++i)
+    values_[i] = a * values_[i] + y_data[i];
```
zachhardy, line 244: See comment above about `std::transform`.

framework/math/ParallelVector/ParallelSTLVector.cc
```diff
@@ -124,104 +199,61 @@ void ParallelSTLVector::operator+=(const ParallelVector& y)
...
-  ChiInvalidArgumentIf(
-    local_end > local_size_,
-    "local_offset + num_values=" + std::to_string(local_end) +
-      ", is out of range for destination vector with local size " +
-      std::to_string(local_size_));
-
-  for (int64_t i = 0; i < num_values; ++i)
-    values_[local_offset + i] = y[y_offset + i];
+  for (size_t i = 0; i < local_size_; ++i)
+    values_[i] *= a;
```
zachhardy, line 250: See comment above about `std::transform`. I assume a range based loop isn't used here to avoid scaling of ghost values.

framework/math/ParallelVector/ParallelSTLVector.cc
```diff
@@ -124,104 +199,61 @@ void ParallelSTLVector::operator+=(const ParallelVector& y)
...
-  ChiInvalidArgumentIf(
-    local_end > local_size_,
-    "local_offset + num_values=" + std::to_string(local_end) +
-      ", is out of range for destination vector with local size " +
-      std::to_string(local_size_));
-
-  for (int64_t i = 0; i < num_values; ++i)
-    values_[local_offset + i] = y[i];
+  for (size_t i = 0; i < local_size_; ++i)
+    values_[i] += a;
```
zachhardy, line 256: See comment above about `std::transform`

framework/math/ParallelVector/ParallelVector.h
```diff
@@ -110,27 +134,16 @@ class ParallelVector
   virtual void operator+=(const ParallelVector& y) = 0;
 
   /**Adds a vector multiplied by scalar a. Optimized for a=1.0 and -1.0*/
```
zachhardy: Please add the equation to the doc. I would do something like 
```
/**Adds a scaled vector to obtain \f$ \vec{x} = \vec{x} + a \vec{y} \f$. Optimized for a = 1.0 and a = -1.0.*/
```
Alternatively, add a line to the doc and us `\f[ ... \f]`.

framework/math/ParallelVector/ParallelVector.h
```diff
@@ -110,27 +134,16 @@ class ParallelVector
...
-  virtual void PlusAY(double a, const ParallelVector& y) = 0;
+  virtual void PlusAY(const ParallelVector& y, double a) = 0;
 
-  /**Sets the local values of one vector equal to another. The sizes must be
-   * compatible.*/
-  virtual void CopyValues(const ParallelVector& y) = 0;
+  /**Performs x = a*x + y with the current vector being x.*/
```
zachhardy, line 140: See above comment above equation doc.

framework/math/ParallelVector/ParallelVector.h
```diff
@@ -110,27 +134,16 @@ class ParallelVector
...
+  virtual void AXPlusY(double a, const ParallelVector& y) = 0;
 
-  /**Copies a contiguous block of local data (num_values entries) from the
-  * source vector (starting at y_offset) to the
-   * current vector starting at local_offset. */
-  virtual void BlockCopyLocalValues(const ParallelVector& y,
-                                    int64_t y_offset,
-                                    int64_t local_offset,
-                                    int64_t num_values) = 0;
+  /**Scales a vector by a scalar*/
```
zachhardy, line 143: Less necessary here, but consider adding the simple equation for doc consistency.

framework/math/ParallelVector/ParallelVector.h
```diff
@@ -110,27 +134,16 @@ class ParallelVector
...
+  virtual void Scale(double a) = 0;
 
-  /**Copies a contiguous block of data from the source STL vector to the
-   * current vector starting at local_offset. The input STL vector must have
-   * exactly num_values entries.
-   * */
-  virtual void BlockCopyLocalValues(const std::vector<double>& y,
-                                    int64_t local_offset,
-                                    int64_t num_values) = 0;
+  /**Adds a constant scalar value to all the entries of the vector*/
```
zachhardy, line 146: A little more necessary here than for Scale, but still not really. Consider adding the simple equation to the doc.

framework/physics/SolverBase/chi_solver.cc
```diff
@@ -22,6 +22,8 @@ chi::InputParameters Solver::GetInputParameters()
 
   params.AddOptionalParameter("dt", 0.01, "Desired initial timestep size.");
   params.AddOptionalParameter("time", 0.0, "Current time of the solver.");
+  params.AddOptionalParameter(
+    "start_time", 1.0, "Transient start-time if applicable.");
```
zachhardy: What is the logic with this default value being the same as `end_time`?

framework/physics/TimeStepControllers/TimeStepper.cc
```diff
@@ -0,0 +1,132 @@
...
+
+chi::InputParameters TimeStepper::GetInputParameters()
+{
+  chi::InputParameters params = ChiObject::GetInputParameters();
+
+  params.AddOptionalParameter("dt", 0.01, "Initial timestep to use");
+  params.AddOptionalParameter("time", 0.0, "Initial time");
+  params.AddOptionalParameter(
+    "time_index", 0, "Time index. Useful for output control.");
+  params.AddOptionalParameter("start_time", 0.0, "Start time");
```
zachhardy: This `start_time` is defaulted to 0.0, the other to 1.0. Is there a reason for this?

framework/physics/TimeStepControllers/TimeStepper.h
```diff
@@ -0,0 +1,85 @@
...
+  /**Manually set the maximum number of time steps. A negative number disables
+* this check.*/
+  void SetMaxTimeSteps(int n);
+
+  /**Advances the controller's state. The most basic action here is to
+  * advance time and the time index.*/
+  virtual void Advance();
+
+  /**Adapts according to the timestep status.*/
+  virtual bool Adapt(TimeStepStatus time_step_status) {return false;}
```
zachhardy: Is this a function to return whether or not to adapt the time step? The function name and doc doesn't scream that to me, but rather that the function computes the adaptive time step. Maybe `DoAdapt` would be better with doc that says `Returns whether time step adaption should be performed.`.

zachhardy (approved):
Most comments are for the same thing. A few recommendations for doc (LaTeX equations), a potential unintentional default parameter, and a function name change recommendation.

framework/math/ParallelVector/ParallelSTLVector.cc
```diff
@@ -0,0 +1,479 @@
+#include "ParallelSTLVector.h"
```
wdhawkins, line 1: Can someone give me the 5 sentence justification for ParallelSTLVector?  Why do we need this vs. the PETSC implementation? It seems to me that maintaining this is going to require a lot of work, and it's likely we won't see the same level of performance as we would using PETSC or another external package.

framework/math/ParallelVector/ParallelSTLVector.cc
```diff
@@ -0,0 +1,479 @@
...
+
+#include "chi_log.h"
+#include "chi_log_exceptions.h"
+
+#include <sstream>
+#include <stdexcept>
+#include <cmath>
+#include <numeric>
+
+#define scint64_t static_cast<int64_t>
```
wdhawkins, line 14: If we need to do this, it suggests to me that there are some fundamental design issues here.

I also see lots of loops and loop operations that are probably better implemented with std::copy and/or std::transform.  Those algorithms may provide better performance than regular loops (it shouldn't be worse anyway) and the code is cleaner.

framework/math/ParallelVector/ParallelSTLVector.cc
```diff
@@ -0,0 +1,479 @@
...
+      else
+        values_[local_id] += value;
+    }
+  }
+
+  // Finally, clear the operation cache
+  op_cache.clear();
+}
+
+std::vector<uint64_t> ParallelSTLVector::DefineExtents(uint64_t local_size,
```
wdhawkins, line 434: Why does this require the communicator be passed in?

wdhawkins (approved):

framework/math/ParallelVector/ParallelSTLVector.cc
zachhardy: There are a few different reasons I originally implemented this. For some background, in @Naktakala implementation of the source corrected DSA k-eigenvalue solver, he put in an option to do PWLC diffusion. This required access to PWLD ghost DoFs. To do that, he created a ghost communicator that was stored within the solver, and then when he needed the PWLD ghosts, he created a local vector, and then used the communicator to populate the ghosts at the back and communicate. I needed similar functionality for a second moment method solver that will be put in the code base in the future. Here two reasons for this:

1. The GhostedParallelSTLVector encapsulated the vector ghost communicator and the actual data in one structure which made the task of getting PWLD ghosts a little more streamlined (and makes the code a bit easier to read). In addition to the PWLD -> PWLC mapping, this was needed for accessing nonlocal tensors for the second moment method.
2. The ParallelSTLVector was a natural thing to implement as a parent of the ghosted version, but how this is currently used is to construct a source vector for a diffusion solver outside of the solver that needs to contribute to nonlocal entries. This source then gets added into the diffusion solver PETSc source vector with the existing routines for adding an STL vector to the diffusion source vector.
3. This could be a platform to have a common interface for vector types from a variety of different packages.

We could certainly use PETSc vectors throughout the entire library if we wanted, so these implementations do nothing other than provide a nicety for a user. It also served as a good learning opportunity for me to get my feet wet in the weeds of this stuff (with lots of help from @Naktakala).

ragusa: @zachhardy I do not fully understand this, especially in light of your last paragraph. My opinion is that we should let PETSc do what it is good at (and its team of developers). Could this have bee a nicer user interface that simply wraps around existing PETSc functionalities? Are we just bloating the base code?

wdhawkins: I agree with @ragusa.  This is my biggest concern.  I could understand and get behind something along the lines of (3), but I'm concerned we will end up with something that requires a great deal of effort to maintain/develop and that just replicates functionality we can get from other packages like PETSC.

framework/physics/TimeStepControllers/ConstantTimeStepper.cc
Naktakala: Yes! and the folder name.

framework/math/PETScUtils/petsc_utils_04_vecmanip.cc
Naktakala: Done

framework/math/PETScUtils/petsc_utils_04_vecmanip.cc
Naktakala: Done

framework/math/ParallelVector/ParallelSTLVector.cc
Naktakala: Done

framework/math/ParallelVector/ParallelSTLVector.cc
Naktakala: Done

framework/math/ParallelVector/ParallelSTLVector.cc
Naktakala: Done

framework/math/ParallelVector/ParallelSTLVector.cc
Naktakala: Done

framework/math/ParallelVector/ParallelSTLVector.cc
Naktakala: I am going to defer this till I can create benchmarks. My experience has always been that the transform via a lambda is pretty cache-unfriendly.

framework/math/ParallelVector/ParallelVector.h
Naktakala: Ooof. The formulas look nasty in the IDE even though they might look cool in html-docs.
image

I think I prefer just having it bare bones. Then it is readable in both html and IDEs.
image

Naktakala:
```
/**Adds a vector multiplied by scalar a, `x = x + a*y`. Optimized for a=1.0 and
   * -1.0*/
  void PlusAY(const ParallelVector& y, double a) override;
```

image

framework/physics/SolverBase/chi_solver.cc
Naktakala: Good catch, Thanks

framework/physics/TimeStepControllers/TimeStepper.h
Naktakala: Changed to:
```
/**Adapts according to the timestep status. If it could provide a change
  * it returns true, otherwise false.*/
  virtual bool Adapt(TimeStepStatus time_step_status) {return false;}
```

framework/math/ParallelVector/ParallelSTLVector.cc
zachhardy: @ragusa The GhostedParallelSTLVector improved upon existing code in ChiTech for ghosted STL vectors. The ParallelSTLVector was essentially given for free in implementing the prior.

I understand the concern here, but what is the logical end of that argument? You could use PETSc for almost everything in the code base, yet we use STL vectors and transfer to and from PETSc vectors when we are about to use a PETSc solver where we could just use PETSc vectors in the first place.

Naktakala: Oh is this an actual serious question?? Have any of you actually used a FEM library where they didn't do this. MOOSE has it, deal.ii has it....

Any c++ developer that's been using PETSc for a while will complain about the following:

PETSc can get quite hectic! 
- It has open/close notions on vectors. 
- The `Vec` and `Mat` are actually typedef on pointers so ownership is never clear. You constantly have to worry about where you can call VecDestroy and MatDestroy
- Cannot use stuff like unique_ptr on them.
- Reading from them is always like 3 lines.
- The headers are absolutely gigantic. Just to get a .cc function to complete the type for Vec you need the petsc.h header, then that absolutely blows up the compile time.
- Ever tried their block vectors? Nightmare.

Geez yeah. No the abstraction in `ParallelVector` and the PETSc-free `ParallelSTLVector` is super valuable. It will benefit our successors 100%. I would waaaay rather transform a PETSc vector straight out of their solvers (to my own), then do my operations on a more c++-fashion vector, until I am ready for PETSc to do its thing, then hand it back. This way PETSc still does the heavy lifting. Only dip into the chaos (efficient chaos yes) momentarily and maintain more sane code outside of PETSc.

ragusa: Wow. Bot from X?

framework/math/ParallelVector/ParallelSTLVector.cc
Naktakala: Good question. It's a static method.

framework/math/ParallelVector/ParallelSTLVector.cc
zachhardy: @ragusa ?

Naktakala: Bot from X????

framework/math/ParallelVector/ParallelSTLVector.cc
wdhawkins: Ah yes... I see.  And why is it static?

framework/math/ParallelVector/ParallelSTLVector.cc
Naktakala: Its pretty simple actually. DOF-IDs are signed integers, 64bit. It is used 3 times in this file. Once because std::copy needs a "difference"-type. Another two more times to convert unsigned sizes to signed so that dof-ids can be computed.

framework/math/ParallelVector/ParallelSTLVector.cc
wdhawkins: I have actually used an FEM library... every day in fact. But not PETSC. I'm not a big fan, and this just reinforces that. This seems like a tremendous amount of effort to work around PETSC's shortcomings.

framework/math/ParallelVector/ParallelSTLVector.cc
wdhawkins: Got it. I actually didn't count the use cases.  3 times is nothing... I was afraid there were numerous uses.  In that case,  it might have been worth it to investigate the target type.

PR #379, opened 2023-09-28
Added obj-reader vertex id check and quality check

Naktakala:
An error in a mesh-file resulted in some strange behavior. The cause was zero length faces so here we've added some quality checks.

### DIFF
--- a/framework/mesh/UnpartitionedMesh/unpartmesh_00a_general.cc
+++ b/framework/mesh/UnpartitionedMesh/unpartmesh_00a_general.cc
@@ -22,7 +22,7 @@ void chi_mesh::UnpartitionedMesh::ComputeCentroidsAndCheckQuality()
 {
   const chi_mesh::Vector3 khat(0.0, 0.0, 1.0);
 
-  Chi::log.Log() << "Computing cell-centroids.";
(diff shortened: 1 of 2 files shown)


framework/mesh/UnpartitionedMesh/unpartmesh_00a_general.cc
```diff
@@ -90,6 +90,50 @@ void chi_mesh::UnpartitionedMesh::ComputeCentroidsAndCheckQuality()
...
+
+        ++f;
+      }
+    } // if polyhedron
+    ++cell_id;
+  } // for cell in raw_cells
+
   if (num_negative_volume_elements > 0)
     Chi::log.LogAllWarning()
       << "Cell quality checks detected " << num_negative_volume_elements
```
ragusa, line 139: can this error message be more explicit by saying negative volumes. this would help the user focus on clockwise/ccw issues in their mesh

ragusa (approved):

framework/mesh/UnpartitionedMesh/unpartmesh_00a_general.cc
ragusa: was my comment unhelpful ?

PR #382, opened 2023-09-30
KBATweak and SPDS fix

Naktakala:
Actually a small bug in SPDS. KBA partitioner should do the same as what PETSc partitioner does, which is to do the partitioning on the home location and then broadcast it back to the others.

### DIFF
--- a/framework/graphs/KBAGraphPartitioner.cc
+++ b/framework/graphs/KBAGraphPartitioner.cc
@@ -91,38 +91,49 @@ KBAGraphPartitioner::Partition(const std::vector<std::vector<uint64_t>>& graph,
     "Graph number of entries not equal to centroids' number of entries.");
   const size_t num_cells = graph.size();
   std::vector<int64_t> pids(num_cells, 0);
-  for (size_t c = 0; c < num_cells; ++c)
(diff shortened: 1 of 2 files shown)


ragusa (approved):

ragusa:
@Naktakala I find it strange that we had some much luck in the past and didn't have a mesh/partition that caused this error.

Naktakala:
> @Naktakala I find it strange that we had some much luck in the past and didn't have a mesh/partition that caused this error.

YES! It is actually very strange that all the combinations came together. 1) Your test quadrature had 4 azimuthal angles per octant, and the pins had 4 segments per octant, this placed angles parallel to many faces in this mesh. 2) Your seemingly random choice of x,y=11, -11 was exactly on a bunch of centroids (like what are the odds).

Any case, thank you for finding it!

ragusa:
@Naktakala OMG ! Product quad and spiderwebs with sectors. We were bound to find that. Thanks so much for the quick turnaround

Every Friday, I am having chi-tech sessions with 5-7 students. The more users we have, the more corner cases we will find. Hopefully, they'll come up to speed soon and will help further test the code

PR #383, opened 2023-10-01
Clean up of chi_math::SpatialDiscretization

Naktakala:
## Highlights
- This PR introduces many clean ups of the spatial discretization system that
  will support better extension in future.
- A lot of functionality has been moved to base classes.
- A lot of documentation was added.
- Completely removed the `SetupFlags` enum as well as the now redundant ability
  to pre-compute and store unit-integrals and quadrature information


modules/DiffusionSolver/Solver/assemble_pwlc.cc
```diff
@@ -2,13 +2,14 @@
...
+// ###################################################################
 /**Assembles PWLC matrix for general cells.*/
 void chi_diffusion::Solver::CFEM_Assemble_A_and_b(chi_mesh::Cell& cell,
                                                   int group)
 {
-  auto pwl_sdm = std::static_pointer_cast<chi_math::SpatialDiscretization_PWLC>(this->discretization_);
-  const auto& fe_intgrl_values = pwl_sdm->GetUnitIntegrals(cell);
+  auto pwl_sdm = std::static_pointer_cast<
+    chi_math::spatial_discretization::PieceWiseLinearContinuous>(this->discretization_);
+  const auto& fe_intgrl_values = unit_integrals_.at(cell.global_id_);
```
ragusa, line 12: likely silly questions. please bear with me

- before we had cell. now it is cell.global_id_. I suppose that before we were passing the global ID but just called it cell or has chi_mesh::Cell changed? I do not see a change to mesh/MeshContinuum/chi_meshcontinuum.h in this PR

- before, we get the unit integrals from pwl_sdm-> . Where is unit_integrals_ in the new version

ragusa (commented):
Are there performance improvements/losses both a low core counts like 8 and at higher core counts like 100 or above?

modules/DiffusionSolver/Solver/assemble_pwlc.cc
Naktakala: Interesting observation. Note: The `DiffusionSolver` module will soon be removed. It is VERY old and has lots of bad practices in it. Here is the gist of what you are seeing.

Units integrals are no longer stored on the spatial discretizations (SDs). We used to be able to ask the SDs to precompute unit integrals via passing `SetupFlags` at creation. We obtained the unit integrals by pass `chi_mesh::Cell&`, which is why you used to see `cell`.

Since the precomputing of unit integrals have now been removed, it is up to the specific solver to compute the ones they need during initialization (LBS already did this). So now, the DiffusionSolver has its own UnitIntegral struct that it assembles for each cell during initialization. Its stores this as a map(global-id->Data) (in order to supported local and ghost cells), that is why you now see `cell.global_id`.

Asking the SDs for unit integrals has systematically been removed until only the legacy `DiffusionSolver` used it. Finally we can now cull it.

Naktakala:
> Are there performance improvements/losses both a low core counts like 8 and at higher core counts like 100 or above?

The nature of this change is such that it is unlikely to impact performance. I have one test I ran at 768 cores and that one showed slight improvement during initialization but not anything drastic (3.5 seconds versus 3.7 seconds).

framework/math/SpatialDiscretization/CellMapping.h
```diff
@@ -0,0 +1,131 @@
...
+   * face-areas. If simple linear cells are used then the
+   * default CellMapping::ComputeCellVolumeAndAreas can be
+   * used as this function. Otherwise (i.e. for higher order
+   * elements, the child-class should
+   * bind a different function to this.*/
+  typedef std::function<void(const chi_mesh::MeshContinuum&,
+                             const chi_mesh::Cell&,
+                             double&,
+                             std::vector<double>&)>
+    VandAFunction;
```
zachhardy: I'm not sure I understand the purpose of having this and passing function handles for an internal static method. Why not just have a mapping call its routine? Why make it static? Why not just the volume and areas in the class routine?

framework/math/SpatialDiscretization/CellMapping.h
```diff
@@ -0,0 +1,131 @@
...
+                                        const chi_mesh::Cell& cell,
+                                        double& volume,
+                                        std::vector<double>& areas);
+
+  const chi_mesh::MeshContinuum& ref_grid_;
+  const chi_mesh::Cell& cell_;
+  const size_t num_nodes_;
+  const std::vector<chi_mesh::Vector3> node_locations_;
+  double volume_ = 0.0;
+  std::vector<double> areas_;
```
zachhardy:
```suggestion
  const chi_mesh::MeshContinuum& grid_;
  const chi_mesh::Cell& cell_;
  
  const size_t num_nodes_;
  const std::vector<chi_mesh::Vector3> node_locations_;
  
  double volume_ = 0.0;
  std::vector<double> areas_;
```

framework/math/SpatialDiscretization/CellMapping.h
```diff
@@ -0,0 +1,131 @@
...
+  GradShapeValues(const chi_mesh::Vector3& xyz,
+                  std::vector<chi_mesh::Vector3>& gradshape_values) const = 0;
+
+  /**Returns the node locations associated with this element.*/
+  const std::vector<chi_mesh::Vector3>& GetNodeLocations() const;
+
+  // 03 Quadrature
+  /**Makes the volumetric/internal quadrature point data for this element.*/
+  virtual finite_element::InternalQuadraturePointData
+  MakeInternalQuadraturePointData() const = 0;
```
zachhardy: I would call this VolumetricQuadraturePointData, or even CellQuadraturePointData. I think that is more familiar to most.

framework/math/SpatialDiscretization/CellMappings/PieceWiseLinear/PieceWiseLinearPolygonMapping.h
```diff
@@ -0,0 +1,84 @@
...
+                         bool on_surface = false);
+
+  // ############################################### Shape functions per side
+  double SideShape(uint32_t side,
+                   uint32_t i,
+                   const chi_mesh::Vector3& qpoint,
+                   bool on_surface = false) const;
+
+  double SideGradShape_x(uint32_t side, uint32_t i) const;
+  double SideGradShape_y(uint32_t side, uint32_t i) const;
```
zachhardy, line 32: Why not `chi_mesh::Vector3 SideGradShape`?

framework/math/SpatialDiscretization/CellMappings/PieceWiseLinear/PieceWiseLinearPolygonMapping.h
```diff
@@ -0,0 +1,84 @@
...
+    double detJ;
+    double detJ_surf;
+    std::array<uint64_t, 2> v_index;
+    chi_mesh::Vector3 v0;
+    chi_mesh::Matrix3x3 J;
+    chi_mesh::Matrix3x3 Jinv;
+    chi_mesh::Matrix3x3 JTinv;
+    chi_mesh::Vector3 normal;
+  };
+  // Goes into sides
```
zachhardy: Confusing placement

framework/math/SpatialDiscretization/CellMappings/PieceWiseLinear/PieceWiseLinearPolygonMapping.h
zachhardy: These routines are private for tets.

framework/math/SpatialDiscretization/CellMappings/PieceWiseLinear/PieceWiseLinearPolyhedronMapping.h
```diff
@@ -0,0 +1,134 @@
...
+  //                                                    tetrahedron linear shape
+  //                                                    functions
+  // 01a_reftet.cc
+  static double TetShape(uint32_t index,
+                         const chi_mesh::Vector3& qpoint,
+                         bool on_surface = false);
+
+  static double TetGradShape_x(uint32_t index);
+  static double TetGradShape_y(uint32_t index);
+  static double TetGradShape_z(uint32_t index);
```
zachhardy, line 56: Same comment as for tris. Why not `chi_mesh::Vector3 TetGradShape`?

zachhardy (approved):
A few notes:

- There seemed to be some mismatched routine scopes in the mappings (i.e. the routines to evaluate shape functions). Check these to make sure they are consistent.
- It would be good to use analogous naming for QP data. I would go with either Volume/Surface or Cell/Face.

framework/math/SpatialDiscretization/CellMappings/PieceWiseLinear/PieceWiseLinearPolygonMapping.h
Naktakala: I agree. I am just going to leave this for future scope since it will require quite some testing

framework/math/SpatialDiscretization/CellMappings/PieceWiseLinear/PieceWiseLinearPolygonMapping.h
Naktakala: VERY old code. Changed to this:
```
// This structure goes into sides
  struct FEside_data2d
  {
    double detJ;
    double detJ_surf;
    std::array<uint64_t, 2> v_index;
    chi_mesh::Vector3 v0;
```

zachhardy: If you want, there is a similar structure for tets too that could be upgraded... However, it made more sense than this does.

framework/math/SpatialDiscretization/CellMappings/PieceWiseLinear/PieceWiseLinearPolygonMapping.h
Naktakala: Good! Done

framework/math/SpatialDiscretization/CellMappings/PieceWiseLinear/PieceWiseLinearPolyhedronMapping.h
Naktakala: Will do in future scope. Would require careful testing

framework/math/SpatialDiscretization/CellMapping.h
Naktakala: As discussed. I will find a replacement for this.

framework/math/SpatialDiscretization/CellMapping.h
Naktakala: Done

PR #387, opened 2023-10-07
Addition of SplitFileMeshGenerator

Naktakala:
This temporarily addresses the issue with the high memory consumption when building meshes.

The `SplitFileMeshGenerator` is very much like MOOSE's capability, just I think a little easier to use since your simulation does not have to turn it on or off all the time.

It essentially builds the mesh only on location 0 then partitions it and writes the necessary pieces to chunks of binary files.

### DIFF
--- a/framework/mesh/MeshGenerator/ExtruderMeshGenerator.cc
+++ b/framework/mesh/MeshGenerator/ExtruderMeshGenerator.cc
@@ -17,9 +17,9 @@ chi::InputParameters ExtrusionLayer::GetInputParameters()
 
   // clang-format off
   params.SetGeneralDescription(
-"A collection of parameters defining an extrusion layer.");
+  "A collection of parameters defining an extrusion layer.");
   // clang-format on
-  params.SetDocGroup("chi_mesh__ExtruderMeshGenerator");
+  params.SetDocGroup("doc_MeshGenerators");
 
   params.AddOptionalParameter(
     "h", 1.0, "Layer height. Cannot be specified if \"z\" is specified.");
@@ -126,6 +126,7 @@ ExtruderMeshGenerator::GenerateUnpartitionedMesh(
   const auto& template_cells = input_umesh->GetRawCells();
 
   const size_t num_template_vertices = template_vertices.size();
+  const size_t num_template_cells = template_cells.size();
 
   ChiLogicalErrorIf(template_vertices.empty(), "Input mesh has no vertices.");
   ChiLogicalErrorIf(template_cells.empty(), "Input mesh has no cells.");
@@ -188,6 +189,7 @@ ExtruderMeshGenerator::GenerateUnpartitionedMesh(
   {
     for (uint32_t n = 0; n < layer.num_sub_layers_; ++n)
     {
+      size_t tc_counter = 0;
       for (const auto& template_cell : template_cells)
       {
         //================================== Determine cell sub-type
@@ -230,7 +232,16 @@ ExtruderMeshGenerator::GenerateUnpartitionedMesh(
           new_face.vertex_ids[3] = tc_face.vertex_ids[0] + (k + 1) * num_template_vertices;
           // clang-format on
 
-          if (not tc_face.has_neighbor) new_face.neighbor = tc_face.neighbor;
+          if (tc_face.has_neighbor)
+          {
+            new_face.neighbor = num_template_cells * k + tc_face.neighbor;
+            new_face.has_neighbor = true;
+          }
+          else
+          {
+            new_face.neighbor = tc_face.neighbor;
+            new_face.has_neighbor = false;
+          }
 
           new_cell.faces.push_back(std::move(new_face));
         } // for tc face
@@ -245,10 +256,20 @@ ExtruderMeshGenerator::GenerateUnpartitionedMesh(
             new_face.vertex_ids.push_back(vid +
                                           (k + 1) * num_template_vertices);
 
-          if ((k + 1) == z_levels.size()) new_face.neighbor = zmax_bndry_id;
+          if (k == (z_levels.size()-2))
+          {
+            new_face.neighbor = zmax_bndry_id;
+            new_face.has_neighbor = false;
+          }
+          else
+          {
+            new_face.neighbor = num_template_cells * (k+1) + tc_counter;
+            new_face.has_neighbor = true;
+          }
 
           new_cell.faces.push_back(std::move(new_face));
         }
+
         // Bottom face
         {
           UnpartitionedMesh::LightWeightFace new_face;
@@ -258,11 +279,22 @@ ExtruderMeshGenerator::GenerateUnpartitionedMesh(
           for (auto vid = vs.rbegin(); vid != vs.rend(); ++vid)
             new_face.vertex_ids.push_back((*vid) + k * num_template_vertices);
 
-          if (k == 0) new_face.neighbor = zmin_bndry_id;
+          if (k == 0)
+          {
+            new_face.neighbor = zmin_bndry_id;
+            new_face.has_neighbor = false;
+          }
+          else
+          {
+            new_face.neighbor = num_template_cells * (k-1) + tc_counter;
+            new_face.has_neighbor = true;
+          }
 
           new_cell.faces.push_back(std::move(new_face));
         }
         umesh->GetRawCells().push_back(new_cell_ptr);
+
+        ++tc_counter;
       } // for template cell
       ++k;
     } // for sub-layer n
--- a/framework/mesh/MeshGenerator/FromFileMeshGenerator.cc
+++ b/framework/mesh/MeshGenerator/FromFileMeshGenerator.cc
@@ -20,7 +20,7 @@ chi::InputParameters FromFileMeshGenerator::GetInputParameters()
 
   params.SetGeneralDescription("Generator for loading an unpartitioned mesh"
                                " from a file.");
-  params.SetDocGroup("MeshGenerator");
+  params.SetDocGroup("doc_MeshGenerators");
 
   params.AddRequiredParameter<std::string>("filename", "Path to the file.");
   params.AddOptionalParameter(
--- a/framework/mesh/MeshGenerator/MeshGenerator.cc
+++ b/framework/mesh/MeshGenerator/MeshGenerator.cc
@@ -104,7 +104,7 @@ void MeshGenerator::Execute()
   if (Chi::mpi.location_id == 0)
     cell_pids = PartitionMesh(*current_umesh, Chi::mpi.process_count);
 
-  BroadCastPIDs(cell_pids, 0, Chi::mpi.comm);
+  BroadcastPIDs(cell_pids, 0, Chi::mpi.comm);
 
   auto grid_ptr = SetupMesh(std::move(current_umesh), cell_pids);
 
@@ -121,15 +121,15 @@ void MeshGenerator::Execute()
   Chi::mpi.Barrier();
 }
 
-void MeshGenerator::SetGridAttributes(chi_mesh::MeshContinuum& grid,
-                                      MeshAttributes new_attribs,
-                                      std::array<size_t, 3> ortho_Nis)
+void MeshGenerator::SetGridAttributes(
+  chi_mesh::MeshContinuum& grid,
+  MeshAttributes new_attribs,
+  std::array<size_t, 3> ortho_cells_per_dimension)
 {
-  grid.SetAttributes(new_attribs, ortho_Nis);
+  grid.SetAttributes(new_attribs, ortho_cells_per_dimension);
 }
 
-void MeshGenerator::ComputeAndPrintStats(
-  const chi_mesh::MeshContinuum& grid)
+void MeshGenerator::ComputeAndPrintStats(const chi_mesh::MeshContinuum& grid)
 {
   const size_t num_local_cells = grid.local_cells.size();
   size_t num_global_cells = 0;
--- a/framework/mesh/MeshGenerator/MeshGenerator.cc
+++ b/framework/mesh/MeshGenerator/MeshGenerator.cc
@@ -3,13 +3,14 @@
 #include "mesh/UnpartitionedMesh/chi_unpartitioned_mesh.h"
 #include "mesh/MeshHandler/chi_meshhandler.h"
 #include "mesh/VolumeMesher/chi_volumemesher.h"
+#include "mesh/MeshContinuum/chi_meshcontinuum.h"
 #include "graphs/GraphPartitioner.h"
 #include "graphs/PETScGraphPartitioner.h"
 
 #include "ChiObjectFactory.h"
 
 #include "chi_runtime.h"
-#include "chi_log.h" // TODO: Remove
+#include "chi_log.h"
 
 namespace chi_mesh
 {
@@ -21,7 +22,7 @@ chi::InputParameters MeshGenerator::GetInputParameters()
   chi::InputParameters params = ChiObject::GetInputParameters();
 
   params.SetGeneralDescription("The base class for all mesh generators");
-  params.SetDocGroup("MeshGenerator");
+  params.SetDocGroup("doc_MeshGenerators");
 
   params.AddOptionalParameter(
     "scale", 1.0, "Uniform scale to apply to the mesh after reading.");
@@ -98,7 +99,14 @@ void MeshGenerator::Execute()
 
   //======================================== Generate final umesh and convert it
   current_umesh = GenerateUnpartitionedMesh(std::move(current_umesh));
-  auto grid_ptr = SetupMesh(std::move(current_umesh));
+
+  std::vector<int64_t> cell_pids;
+  if (Chi::mpi.location_id == 0)
+    cell_pids = PartitionMesh(*current_umesh, Chi::mpi.process_count);
+
+  BroadCastPIDs(cell_pids, 0, Chi::mpi.comm);
+
+  auto grid_ptr = SetupMesh(std::move(current_umesh), cell_pids);
 
   //======================================== Assign the mesh to a VolumeMesher
   auto new_mesher =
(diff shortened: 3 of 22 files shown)


framework/mesh/MeshGenerator/ExtruderMeshGenerator.cc
zachhardy, line 1: General question: Should this be `Extruded` instead of `Extruder`?

framework/mesh/MeshGenerator/ExtruderMeshGenerator.cc
```diff
@@ -17,9 +17,9 @@ chi::InputParameters ExtrusionLayer::GetInputParameters()
...
   // clang-format off
   params.SetGeneralDescription(
-"A collection of parameters defining an extrusion layer.");
+  "A collection of parameters defining an extrusion layer.");
   // clang-format on
-  params.SetDocGroup("chi_mesh__ExtruderMeshGenerator");
+  params.SetDocGroup("doc_MeshGenerators");
 
   params.AddOptionalParameter(
     "h", 1.0, "Layer height. Cannot be specified if \"z\" is specified.");
```
zachhardy, line 25: It is unclear what "if \"z\" specified" means here since there is no \"z\" argument. Consider rephrasing.

framework/mesh/MeshGenerator/FromFileMeshGenerator.cc
zachhardy, line 1: Is this class a base class for reading from various different file types? If so, all good. If not, what file type does this read and why is it not specified?

framework/mesh/MeshGenerator/MeshGenerator.cc
```diff
@@ -113,4 +121,70 @@ void MeshGenerator::Execute()
   Chi::mpi.Barrier();
 }
 
+void MeshGenerator::SetGridAttributes(chi_mesh::MeshContinuum& grid,
+                                      MeshAttributes new_attribs,
+                                      std::array<size_t, 3> ortho_Nis)
```
zachhardy: A little out of scope, but `ortho_Nis` is an ambiguous variable name. I think it means number of orthogonal cells per dimension (i.e. Nx, Ny, Nz), but `Nis` is ambiguous. `ortho_cells/nodes_per_dim` would be clear (not sure if `Nis` is cells or nodes.

framework/mesh/MeshGenerator/MeshGenerator.h
```diff
@@ -40,30 +40,63 @@ class MeshGenerator : public ChiObject
...
+    };
+    const T& list_;
+  };
+
+protected:
   // 01
+  /**Builds a cell-graph and executes the partitioner that assigns cell
+   * partition ids based on the supplied number of parts.*/
+  std::vector<int64_t> PartitionMesh(const UnpartitionedMesh& input_umesh,
+                                     int num_parts);
```
zachhardy:
```suggestion
  /**Builds a cell-graph and executes the partitioner that assigns cell
   * partition ids based on the supplied number of partitions.*/
  std::vector<int64_t> PartitionMesh(const UnpartitionedMesh& input_umesh,
                                     int num_partitions);
```

framework/mesh/MeshGenerator/MeshGenerator.h
```diff
@@ -40,30 +40,63 @@ class MeshGenerator : public ChiObject
...
-  virtual std::shared_ptr<MeshContinuum>
-  SetupMesh(std::unique_ptr<UnpartitionedMesh> input_umesh_ptr);
+  std::shared_ptr<MeshContinuum>
+  SetupMesh(std::unique_ptr<UnpartitionedMesh> input_umesh_ptr,
+            const std::vector<int64_t>& cell_pids);
 
   // 02 utils
+  /**Broadcasts PIDs to other locations.*/
+
+  static void BroadCastPIDs(std::vector<int64_t>& cell_pids,
```
zachhardy:
```suggestion
  static void BroadcastPIDs(std::vector<int64_t>& cell_pids,
```

framework/mesh/MeshGenerator/MeshGenerator.h
```diff
@@ -40,30 +40,63 @@ class MeshGenerator : public ChiObject
...
+                            int root,
+                            MPI_Comm communicator);
   /**Determines if a cells needs to be included as a ghost or as a local cell.*/
   bool
-  CellHasLocalScope(const chi_mesh::UnpartitionedMesh::LightWeightCell& lwcell,
+  CellHasLocalScope(int location_id,
```
zachhardy, line 82: Unless this has a specific meaning `IsCellLocal` would be more clear, in my view.

framework/mesh/MeshGenerator/MeshGenerator_01_SetupMesh.cc
zachhardy, line 1: General comment on file naming: Are we going for this format going forward? I haven't seen CamelCase (I think it is called) anywhere we have a `_0x_` filename before.

framework/mesh/MeshGenerator/MeshGenerator_01_SetupMesh.cc
```diff
@@ -8,11 +8,12 @@
 
 namespace chi_mesh
 {
-/**Executes the partitioner and configures the mesh as a real mesh.*/
-std::shared_ptr<MeshContinuum>
-MeshGenerator::SetupMesh(std::unique_ptr<UnpartitionedMesh> input_umesh_ptr)
+/**Builds a cell-graph and executes the partitioner.*/
+std::vector<int64_t>
+MeshGenerator::PartitionMesh(const UnpartitionedMesh& input_umesh,
+                             int num_parts)
```
zachhardy:
```suggestion
std::vector<int64_t>
MeshGenerator::PartitionMesh(const UnpartitionedMesh& input_umesh,
                             int num_partitions)
```

framework/mesh/MeshGenerator/MeshGenerator_02_utils.cc
```diff
@@ -3,27 +3,54 @@
 namespace chi_mesh
 {
 
+// ###################################################################
+/**Broadcasts PIDs to other locations.*/
+void MeshGenerator::BroadCastPIDs(std::vector<int64_t>& cell_pids,
```
zachhardy:
```suggestion
void MeshGenerator::BroadcastPIDs(std::vector<int64_t>& cell_pids,
```

framework/mesh/MeshGenerator/OrthogonalMeshGenerator.cc
```diff
@@ -48,6 +48,17 @@ OrthogonalMeshGenerator::OrthogonalMeshGenerator(
...
+      node_set.size() < 2,
+      "Node set " + std::to_string(ns) + " only has " +
+        std::to_string(node_set.size()) +
+        " nodes. A minimum of 2 is required to define a cell.");
+    ++ns;
+  }
+
   ChiInvalidArgumentIf(node_sets_.size() > 3,
                        "More than 3 node sets have been provided. Only a "
                        "maximum of 3 are allowed");
```
zachhardy:
```suggestion
  ChiInvalidArgumentIf(node_sets_.size() > 3,
                       "More than 3 node sets have been provided. The "
                       "maximum allowed is 3.");
```
I would move this before the loop over node sets as well. No reason to go through the nodes if this condition is met.

framework/mesh/MeshGenerator/SplitFileMeshGenerator.cc
```diff
@@ -0,0 +1,398 @@
...
+    "Generates the mesh only on location 0, thereafter partitions the mesh"
+    " but instead of broadcasting the mesh to other locations it creates binary"
+    " mesh files for each location.");
+  params.SetDocGroup("doc_MeshGenerators");
+
+  params.AddOptionalParameter("num_parts",
+                              0,
+                              "The number of parts to generate. If zero will "
+                              "default to the number of MPI processes. Is "
+                              "ignored if the number of MPI processes > 1.");
```
zachhardy: Is parts partitions? I think it should be spelt out if it is.

framework/mesh/MeshGenerator/SplitFileMeshGenerator.h
zachhardy, line 1: The naming is a little weird to me. Would `PartitionFileMeshGenerator` be more appropriate?

framework/mesh/MeshGenerator/SplitFileMeshGenerator.h
```diff
@@ -0,0 +1,56 @@
...
+  void Execute() override;
+
+protected:
+  void WriteSplitMesh(const std::vector<int64_t>& cell_pids,
+                      const UnpartitionedMesh& umesh,
+                      int num_parts);
+  static void SerializeCell(const UnpartitionedMesh::LightWeightCell& cell,
+                            chi_data_types::ByteArray& serial_buffer);
+  typedef std::pair<int, uint64_t> CellPIDGID;
+  struct SplitMeshInfo
```
zachhardy, line 32: What about `LocalMeshInfo`?

zachhardy (approved):
Please consider the suggestions from the comments.

framework/mesh/MeshGenerator/ExtruderMeshGenerator.cc
Naktakala: Did the diff perhaps hide some scope:
```
params.AddOptionalParameter(
    "h", 1.0, "Layer height. Cannot be specified if \"z\" is specified.");
  params.AddOptionalParameter("n", 1, "Number of sub-layers");
  params.AddOptionalParameter("z",
                              0.0,
                              "The z-coordinate at the top of the layer. "
                              "Cannot be specified if \"n\" is specified.");
```

framework/mesh/MeshGenerator/FromFileMeshGenerator.cc
Naktakala: Its not a base class but it is generic:
```
  if (extension == ".obj") umesh->ReadFromWavefrontOBJ(options);
  else if (extension == ".msh")
    umesh->ReadFromMsh(options);
  else if (extension == ".e")
    umesh->ReadFromExodus(options);
  else if (extension == ".vtu")
    umesh->ReadFromVTU(options);
  else if (extension == ".pvtu")
    umesh->ReadFromPVTU(options);
  else if (extension == ".case")
    umesh->ReadFromEnsightGold(options);
  else
    ChiInvalidArgument("Unsupported file type \"" + extension +
                       "\". Supported types limited to"
                       ".obj, .msh, .e, .vtu, .pvtu, .case.");

```

framework/mesh/MeshGenerator/MeshGenerator.h
Naktakala: There are local cells and ghost cells. Both have local scope. This routine determines if a cell is either a local cell or a ghost, OR, if the mesh is replicated, then this will always return true.

framework/mesh/MeshGenerator/MeshGenerator.cc
Naktakala: `ortho_cells_per_dimension` done

framework/mesh/MeshGenerator/MeshGenerator.h
Naktakala: Done

framework/mesh/MeshGenerator/MeshGenerator.h
Naktakala: Done

framework/mesh/MeshGenerator/OrthogonalMeshGenerator.cc
Naktakala: Done

framework/mesh/MeshGenerator/SplitFileMeshGenerator.cc
Naktakala: Depends on how you look at it. What's the difference between a partition and a part? IDK. Was just trying to keep input language short, but hey, "death to the user!"-Clue.

Renamed to num_partitions

framework/mesh/MeshGenerator/SplitFileMeshGenerator.h
Naktakala: Negative because the cells and the vertices in the mesh is not exclusively local. They may have ghost relations

framework/mesh/MeshGenerator/ExtruderMeshGenerator.cc
Naktakala: IDK.... I've become accustomed to saying "use the extruder", sort of like a brand name. If you look at it as Extruded-Mesh Generator it makes more sense, but if you say it like Extruder Mesh-Generator then it also makes sense.

framework/mesh/MeshGenerator/SplitFileMeshGenerator.h
Naktakala: What is a `PartitionFile`? Maybe `PartitionedFile` or `SplitFile`

framework/mesh/MeshGenerator/ExtruderMeshGenerator.cc
zachhardy: Well I get it, but it doesn't extend well to other Mesh-Generators unless Ortho, FromFile, SplitFile, etc. are deemed a mesh "brand". I think `<object_name>Generator` is the better convention here. Up to you.

framework/mesh/MeshGenerator/SplitFileMeshGenerator.h
zachhardy: `PartitionFile` is a file for a partition. That was my thinking. `PartitionedFile` would be fine as well as that implies a single mesh file partitioned across multiple processes.

framework/mesh/MeshGenerator/SplitFileMeshGenerator.cc
zachhardy: Well a part could be a control rod or a fuel rod, i.e. like a physical part you are meshing. Thanks.

framework/mesh/MeshGenerator/SplitFileMeshGenerator.h
zachhardy: Hmmm. Since this is a `SplitFileMesh` saying `SplitMeshInfo` sounds like it is a global thing, as opposed to a struct containing the local portion (plus ghosts) of the global mesh. Within the discretizations, we can get a ghost DoF from `MapDOFLocal`. While the ghost DoF is not truly local, there is local access. I would make a similar argument here. A local mesh is all of the information needed locally to fit together with the global mesh. Take this for what you will

wdhawkins (approved):

framework/mesh/MeshGenerator/FromFileMeshGenerator.cc
```diff
@@ -20,7 +20,7 @@ chi::InputParameters FromFileMeshGenerator::GetInputParameters()
 
   params.SetGeneralDescription("Generator for loading an unpartitioned mesh"
                                " from a file.");
-  params.SetDocGroup("MeshGenerator");
+  params.SetDocGroup("doc_MeshGenerators");
```
ragusa, line 23: should this be a more detailed description?

framework/mesh/MeshGenerator/MeshGenerator_02_utils.cc
```diff
@@ -3,27 +3,54 @@
...
 // ###################################################################
 /**Determines if a cells needs to be included as a ghost or as a local cell.*/
-bool chi_mesh::MeshGenerator::CellHasLocalScope(
+bool MeshGenerator::CellHasLocalScope(
+  int location_id,
   const chi_mesh::UnpartitionedMesh::LightWeightCell& lwcell,
   uint64_t cell_global_id,
   const std::vector<std::set<uint64_t>>& vertex_subscriptions,
-  const std::vector<int64_t>& cell_partition_ids)
+  const std::vector<int64_t>& cell_partition_ids) const
```
ragusa, line 40: can you move the ") const" to the next line for readability? I finally understood that the last const applied to the whole method. the ")" threw me in!

framework/mesh/MeshGenerator/MeshGenerator_02_utils.cc
```diff
@@ -3,27 +3,54 @@
...
   int cell_pid = static_cast<int>(cell_partition_ids[cell_global_id]);
-  if (cell_pid == Chi::mpi.location_id) return true;
+  if (cell_pid == location_id) return true;
 
   // Now determine if the cell is a ghost cell
   for (uint64_t vid : lwcell.vertex_ids)
     for (uint64_t cid : vertex_subscriptions[vid])
     {
       if (cid == cell_global_id) continue;
       int adj_pid = static_cast<int>(cell_partition_ids[cid]);
```
ragusa, line 52: why are we downcasting an int64_t to an int? do we expect more than 2 billion partitions (cell_partition_ids)? and if so (in the CBC sweeper for example), then is it safe to downcast?

framework/mesh/MeshGenerator/MeshGenerator_02_utils.cc
```diff
@@ -32,10 +59,11 @@ bool chi_mesh::MeshGenerator::CellHasLocalScope(
...
 /**Converts a light-weight cell to a real cell.*/
 std::unique_ptr<chi_mesh::Cell>
-MeshGenerator::SetupCell(const UnpartitionedMesh::LightWeightCell& raw_cell,
-                         uint64_t global_id,
-                         uint64_t partition_id,
-                         const std::vector<chi_mesh::Vector3>& vertices)
+MeshGenerator::SetupCell(
+  const UnpartitionedMesh::LightWeightCell& raw_cell,
+  uint64_t global_id,
+  uint64_t partition_id,
```
ragusa, line 65: before, cell_partition_ids was an int64_t. here, partition_id is uint64_t. Why is one unsigned and the other one signed? just curious.

framework/mesh/MeshGenerator/OrthogonalMeshGenerator.cc
```diff
@@ -14,7 +14,7 @@ chi::InputParameters OrthogonalMeshGenerator::GetInputParameters()
   chi::InputParameters params = MeshGenerator::GetInputParameters();
 
   params.SetGeneralDescription("Creates orthogonal meshes.");
-  params.SetDocGroup("MeshGenerator");
+  params.SetDocGroup("doc_MeshGenerators");
```
ragusa, line 17: same Q about the param doc. maybe I do not understand what it should be. is doc_MeshGenerators ?

ragusa (commented):
how do we know that all ranks>0 have not yet allocated too much memory so that rank0 has most of the memory on the node and can do its meshing business?? I suppose the reason is because none of the sweeping data (tdgs, fluds, spuds) cannot be allocated at this poin, so this is why this works, right?

framework/mesh/MeshGenerator/SplitFileMeshGenerator.cc
```diff
@@ -0,0 +1,399 @@
...
+    num_parts_(params.GetParamValue<int>("num_partitions")),
+    split_mesh_dir_path_(
+      params.GetParamValue<std::string>("split_mesh_dir_path")),
+    split_file_prefix_(params.GetParamValue<std::string>("split_file_prefix")),
+    read_only_(params.GetParamValue<bool>("read_only"))
+{
+}
+
+// ##################################################################
+void SplitFileMeshGenerator::Execute()
```
ragusa, line 72: can you point me to the location where I expect that rank-0 will release most of the memory it used for create the mesh and partition it? I am looking for where rank-0 is releasing memory. Thanks!!

framework/mesh/MeshGenerator/FromFileMeshGenerator.cc
Naktakala: So these are tags to documentation pages and is not user facing. In doxygen you create a page like this
```
\** \defgroup <tag_name> Multiword description
\ingroup <another_tag>

Page contents bla bla 
*\
```

To place this object as a child of a group we use that `SetDocGroup` method otherwise the page for its documentation will go in the main category, i.e., `Input Reference`. So there is now better way to keep it short and sweet but to just use tags. Here we link to the page with tag `doc_MeshGenerators`.

ragusa (approved):

framework/mesh/MeshGenerator/MeshGenerator_02_utils.cc
Naktakala: Our coding convention right now, in the .clang-format file, is set up to place it on the same line, which I also believe is the better way to do it because it does save space. It might be that you are not used to working with the const functions (like I was), so can I please ask you to hang in there and see if it gets better. The alternative is to go to openFOAM style and have ever increasing line counts which I don't believe you want either.

framework/mesh/MeshGenerator/MeshGenerator_02_utils.cc
Naktakala: There is an interesting disparity in 3 places:
- We store `partition_id` on a cell as a `uint64_t`
- MPI uses `int` exclusively for ranks and rank count
- Finally, PETSc's partitioners work on `int64_t`

We store the partition_id as `uint64_t` because it simplified serialization and its one less bound to check since it unsigned, however, we could refactor to just `int` it would just require some small changes here and there.

Overall, the `int` on the mpi side is the determining factor. MPI just would start with an overflow of int so it is safe to downcast to it.

framework/mesh/MeshGenerator/MeshGenerator_02_utils.cc
Naktakala: More detail in previous comment but synopsis here: `int64_t` is what the PETSc partitioners operate on and `uint64_t` is what we store on a cell.

framework/mesh/MeshGenerator/OrthogonalMeshGenerator.cc
Naktakala: Tag to the main page for all MeshGenerators.

framework/mesh/MeshGenerator/SplitFileMeshGenerator.cc
Naktakala: Sure. Below is the first part of the `Execute` method:
```c++
void SplitFileMeshGenerator::Execute()
{
  const int num_mpi = Chi::mpi.process_count;
  const int num_parts = num_mpi == 1 ? num_parts_ : num_mpi;

  if (Chi::mpi.location_id == 0 and (not read_only_))
  {
    //======================================== Execute all input generators
    // Note these could be empty
    std::unique_ptr<UnpartitionedMesh> current_umesh = nullptr;
    for (auto mesh_generator_ptr : inputs_)
    {
      auto new_umesh =
        mesh_generator_ptr->GenerateUnpartitionedMesh(std::move(current_umesh));
      current_umesh = std::move(new_umesh);
    }

    //======================================== Generate final umesh
    current_umesh = GenerateUnpartitionedMesh(std::move(current_umesh));

    Chi::log.Log() << "Writing split-mesh with " << num_parts << " parts";
    const auto cell_pids = PartitionMesh(*current_umesh, num_parts);
    WriteSplitMesh(cell_pids, *current_umesh, num_parts);
    Chi::log.Log() << "Split-mesh with " << num_parts
                   << " parts successfully created";
  } // if home location
```

The unique ptr `std::unique_ptr<UnpartitionedMesh> current_umesh` only exists within the scope of the if statement ` if (Chi::mpi.location_id == 0 and (not read_only_))`. The moment we leave that scope the unique_ptr destroys the mesh it was pointing to. Thereafter all processors join up at the next Barrier after which each one reads their piece from the respective file. ;-)

Naktakala:
> how do we know that all ranks>0 have not yet allocated too much memory so that rank0 has most of the memory on the node and can do its meshing business?? I suppose the reason is because none of the sweeping data (tdgs, fluds, spuds) cannot be allocated at this poin, so this is why this works, right?

Yeah, this happens before the solvers are even created. This was the cause of the memory spike. Now procs > 0 waits at a barrier for proc-0 to do its thing

PR #388, opened 2023-10-11
Better timing block

Naktakala:
## Highlights
This is a fairly simple addition. I created the `RepeatingEvent` system a long time a go to visualize sweeps in parallel, however, that ended up being used for sweep timing too. The flaw in that system is that it keeps a history and therefore if one naively used it for timing in very granular codes (like within a cell/face loop) then you would see something that looked like a memory leak because the history would grow to millions.

A much better system is now the hierarchical `chi::TimingBlock` system administered via `chi::TimingLog`, from which `ChiLog` inherits. By default there is a master timing block `ChiTech` which is the default parent for all blocks, however, parents for sub-blocks can be assigned to build a timing or performance graph.

## Example:
This is also a test in the test-system:
```c++
 auto& t_main = Chi::log.CreateTimingBlock("Timing_Main");
  t_main.TimeSectionBegin();
  {
    // Some overhead
    chi::Sleep(std::chrono::milliseconds(200));

    auto& t_1 = Chi::log.CreateOrGetTimingBlock("Part1", "Timing_Main");
    t_1.TimeSectionBegin();
    {
      chi::Sleep(std::chrono::milliseconds(300));
    }
    t_1.TimeSectionEnd();

    auto& t_2 = Chi::log.CreateOrGetTimingBlock("Part2", "Timing_Main");
    t_2.TimeSectionBegin();
    {
      chi::Sleep(std::chrono::milliseconds(333));
    }
    t_2.TimeSectionEnd();

    auto& t_3 = Chi::log.GetTimingBlock("Part2");
    t_3.TimeSectionBegin();
    {
      chi::Sleep(std::chrono::milliseconds(123));
    }
    t_3.TimeSectionEnd();
  }
  t_main.TimeSectionEnd();
```

The graph can be printed via the lua language using `chiLogPrintTimingGraph()` and it looks like this:
```
[0]  *---------------*--------*---------------*-----------------*-------------*
[0]  | Section Name  | #calls | Total time[s] | Average time[s] | % of parent |
[0]  *---------------*--------*---------------*-----------------*-------------*
[0]  | ChiTech       |      1 |        0.9724 |          0.9724 |          -- |
[0]  |   Timing_Main |      1 |       0.97232 |         0.97232 |      99.99% |
[0]  |     Part1     |      1 |       0.30504 |         0.30504 |      31.37% |
[0]  |     Part2     |      2 |        0.4641 |         0.23205 |      47.73% |
[0]  *---------------*--------*---------------*-----------------*-------------*
```

### DIFF
--- a/framework/chi_runtime.cc
+++ b/framework/chi_runtime.cc
@@ -189,6 +189,8 @@ int Chi::Initialize(int argc, char** argv, MPI_Comm communicator)
 
   run_time::InitPetSc(argc, argv);
 
+  auto& t_main = Chi::log.CreateTimingBlock("ChiTech");
+  t_main.TimeSectionBegin();
   chi::SystemWideEventPublisher::GetInstance().PublishEvent(chi::Event(
     "ProgramStart", chi::GetStandardEventCode("ProgramStart")));
 
@@ -213,6 +215,8 @@ int Chi::run_time::InitPetSc(int argc, char** argv)
  * */
 void Chi::Finalize()
 {
+  auto& t_main = Chi::log.GetTimingBlock("ChiTech");
+  t_main.TimeSectionEnd();
   chi::SystemWideEventPublisher::GetInstance().PublishEvent(chi::Event(
     "ProgramExecuted", chi::GetStandardEventCode("ProgramExecuted")));
   meshhandler_stack.clear();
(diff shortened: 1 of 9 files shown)


ragusa (approved):
Nice!

PR #389, opened 2023-10-13
Pqaud and product quadrature fix

ragusa:


Naktakala (approved):
Fairly painless

zachhardy (approved):
Good catch on `pqaud` instead of `pquad` in the tests!

PR #394, opened 2023-10-17
Lagrange spatial discretization finally here

Naktakala:
# Introduction of the Lagrange spatial discretization

As most of you know the PWL discretization treats all 2D cells as polygons 
comprised of sub-triangles and all 3D cells as polyhedra comprised of tetrahedra.
This generally results in an excessive amount of quadrature points when dealing
with standard elements like quads, triangles, hexes and prisms. With the 
introduction of the Lagrange spatial discretizations we can now treat standard
elements as ... standard elements and only activate the PWL stuff when 
non-standard elements are encountered.

Overall this is a standalone change, however, eventually LBS needs to be moved
over to this.

## Highlights
- New cell mappings
- New spatial discretizations
- New spatial weighting function to perform integrals in RZ and 1DSperical
- Two new test meshes
- Hoards of tests for PWL and Lagrange


ragusa:
A first comment: why would lbs need to be moved over? we only compute and store the elemental matrix at the first sweep; after that, we reuse then. so the matrices in lbs need only to be updated if material properties, for instance, change. and even if material properties change, if they change in a cell-wise constant manner, then re-computation is not necessary. I am not saying anything against the Lagrange SD, I am just commenting that for lbs, I am not seeing an immediate need. I also would like the scaling and timing of lbs to be assessed and documented (what Daryl is working on) so that we have a good basis for comparison (at scale) when large changes happen in lbs

framework/math/SpatialDiscretization/CellMappings/Lagrange/LagrangeHexMapping.cc
```diff
@@ -0,0 +1,157 @@
...
+  const double y = qpoint.y;
+  const double z = qpoint.z;
+
+  const double ab = a * b;
+  const double ac = a * c;
+  const double bc = b * c;
+  const double abc = a * bc;
+  return Vec3(0.125 * (a + ab * y + ac * z + abc * y * z),
+              0.125 * (b + ab * x + bc * z + abc * x * z),
+              0.125 * (c + bc * y + ac * x + abc * x * y));
```
ragusa: a for 1st component (and b for 2nd, and c for 3rd) can be factored out. I find the definition more complex because expanded. why not a*(1+b*y)(1+c*z) and similar for other components. it's also more relatable to LagrangeHexMapping::RefShape

framework/math/SpatialDiscretization/CellMappings/Lagrange/LagrangeHexMapping.h
```diff
@@ -0,0 +1,35 @@
...
+class LagrangeHexMapping : public LagrangeBaseMapping
+{
+public:
+  LagrangeHexMapping(const chi_mesh::MeshContinuum& grid,
+                     const chi_mesh::Cell& cell,
+                     const Quadrature& volume_quadrature,
+                     const Quadrature& surface_quadrature);
+
+protected:
+  double RefShape(uint32_t i, const Vec3& qpoint) const override;
```
ragusa, line 20: just for me, what does override to the right do again?

framework/math/SpatialDiscretization/CellMappings/Lagrange/LagrangeWedgeMapping.cc
```diff
@@ -0,0 +1,195 @@
+#include "LagrangeWedgeMapping.h"
```
ragusa, line 1: Nice to have wedges !!

framework/math/SpatialDiscretization/CellMappings/doc/doc_CellMappings.h
```diff
@@ -1,3 +1,10 @@
 /**\defgroup doc_CellMappings Cell Mappings
 *
+In order to for a cell to map to an actual Finite Element we need an interface
```
ragusa: remove "to" maybe

framework/math/SpatialDiscretization/CellMappings/doc/doc_CellMappings.h
```diff
@@ -1,3 +1,10 @@
...
+class that can map a geometric cell to an element. The different types of
+elements in existence is quite vast and here we only have a subset of the most
+popular ones. The nice thing ChiTech is that we support elements beyond just
+the traditional elements, i.e. triangles, quadrilaterals, tetrahedrons and
+hexahedrons, to the general class of polygons and polyhedra.
```
ragusa: tetrahedra 
polyhedra

framework/math/SpatialDiscretization/SpatialDiscretizatoin.cc
ragusa, line 1: there is a typo in the name of that routine. "xxxtoin.cc" --> "xxxtion.cc""

framework/math/SpatialDiscretization/SpatialDiscretizatoin.cc
```diff
@@ -310,14 +310,14 @@ double SpatialDiscretization::CartesianSpatialWeightFunction(
 double SpatialDiscretization::CylindricalRZSpatialWeightFunction(
   const chi_mesh::Vector3& point)
 {
-  return point[0];
+  return 2.0 * M_PI * point[0];
```
ragusa, line 313: is this used inn lbs cylindrical? I think not, we have to have be careful about the convention used there when we look at it later nad possibly change it

test/framework/chi_math/SpatialDiscretization/DG/sdm_test_01_discontinuous.cc
ragusa, line 1: I like that. So the CFEM and DFEM diffusion may go away at some poitn thanks to this, right?

ragusa (approved):

framework/math/SpatialDiscretization/CellMappings/Lagrange/LagrangeHexMapping.cc
Naktakala: Done

framework/math/SpatialDiscretization/CellMappings/doc/doc_CellMappings.h
Naktakala: Done

framework/math/SpatialDiscretization/CellMappings/doc/doc_CellMappings.h
Naktakala: Done

framework/math/SpatialDiscretization/SpatialDiscretizatoin.cc
Naktakala: It does get used there. And it makes no difference since it applies a constant to the whole equation. Where this makes a difference is when one computes volumetric quantities like total power from heat generation rate. The 2pi is necessary to compute the correct volume.

Naktakala: I encountered it in a post-processor

test/framework/chi_math/SpatialDiscretization/DG/sdm_test_01_discontinuous.cc
Naktakala: Yes. I so dearly want that old solver out

framework/math/SpatialDiscretization/CellMappings/LagrangeBaseMapping.h
```diff
@@ -0,0 +1,116 @@
...
+   * on the face index.*/
+  virtual const Quadrature& GetSurfaceQuadrature(size_t face_index) const;
+
+  const Quadrature& volume_quadrature_;
+  const Quadrature& surface_quadrature_;
+};
+
+/**Helper class to apply Newton's method to map world coordinates to
+ * reference element coordinates.*/
+class LagrangeWorldXYZToNaturalMappingHelper : public NonLinearFunction
```
zachhardy: I have always used "global" and "reference" coordinates so "natural" confused me at first. In the doc you say "reference element coordinates" and not "natural". It would be helpful for a user not familiar with the notation to say "reference element (natural) coordinates" in the doc.

The class name also confused me a bit at first because it read "Lagrange World XYZ - To - Natural Mapping - Helper". I would recommend `WorldXYZToNaturalXYZMappingHelper` instead.

framework/math/SpatialDiscretization/CellMappings/LagrangeBaseMapping.cc
```diff
@@ -0,0 +1,323 @@
...
+
+      shape_grad[i][qp] = {JTInv_b[0], JTInv_b[1], JTInv_b[2]};
+
+      const auto& node_xyz = node_locations_[i];
+
+      qpoints_xyz[qp] += ref_shape_i * node_xyz;
+    } // for i
+  }   // for qp
+
+  num_nodes = num_nodes_;
```
zachhardy:
```suggestion
```

framework/math/SpatialDiscretization/CellMappings/LagrangeBaseMapping.cc
```diff
@@ -0,0 +1,323 @@
...
+  typedef std::vector<Vec3> VecVec3;
+  const size_t num_qpoints = volume_quadrature_.qpoints_.size();
+
+  //=================================== Declare necessary vars
+  std::vector<unsigned int> quadrature_point_indices;
+  VecVec3 qpoints_xyz;
+  std::vector<VecDbl> shape_value;
+  std::vector<VecVec3> shape_grad;
+  VecDbl JxW;
+  size_t num_nodes;
```
zachhardy:
```suggestion
```

framework/math/SpatialDiscretization/CellMappings/LagrangeBaseMapping.cc
```diff
@@ -0,0 +1,323 @@
...
+
+  num_nodes = num_nodes_;
+
+  return finite_element::VolumetricQuadraturePointData(quadrature_point_indices,
+                                                     qpoints_xyz,
+                                                     shape_value,
+                                                     shape_grad,
+                                                     JxW,
+                                                     face_node_mappings_,
+                                                     num_nodes);
```
zachhardy:
```suggestion
                                                     num_nodes_);
```

framework/math/SpatialDiscretization/CellMappings/LagrangeBaseMapping.cc
```diff
@@ -0,0 +1,323 @@
...
+
+      shape_grad[i][qp] = {JTInv_b[0], JTInv_b[1], JTInv_b[2]};
+
+      const auto& node_xyz = node_locations_[i];
+
+      qpoints_xyz[qp] += ref_shape_i * node_xyz;
+    } // for i
+  }   // for qp
+
+  num_nodes = num_nodes_;
```
zachhardy:
```suggestion
```

framework/math/SpatialDiscretization/CellMappings/LagrangeBaseMapping.cc
```diff
@@ -0,0 +1,323 @@
...
+  const size_t num_qpoints = surface_quadrature.qpoints_.size();
+
+  //=================================== Declare necessary vars
+  std::vector<unsigned int> quadrature_point_indices;
+  VecVec3 qpoints_xyz;
+  std::vector<VecDbl> shape_value;
+  std::vector<VecVec3> shape_grad;
+  VecDbl JxW;
+  VecVec3 normals;
+  size_t num_nodes;
```
zachhardy:
```suggestion
```

framework/math/SpatialDiscretization/CellMappings/LagrangeBaseMapping.cc
```diff
@@ -0,0 +1,323 @@
...
+  num_nodes = num_nodes_;
+
+  return finite_element::SurfaceQuadraturePointData(quadrature_point_indices,
+                                                 qpoints_xyz,
+                                                 shape_value,
+                                                 shape_grad,
+                                                 JxW,
+                                                 normals,
+                                                 face_node_mappings_,
+                                                 num_nodes);
```
zachhardy:
```suggestion
                                                 num_nodes_);
```

framework/math/SpatialDiscretization/FiniteElement/Lagrange/LagrangeBase.cc
```diff
@@ -0,0 +1,149 @@
...
+    tri_quad_order_arbitrary_(q_order),
+    quad_quad_order_arbitrary_(q_order),
+    tet_quad_order_arbitrary_(q_order),
+    hex_quad_order_arbitrary_(q_order),
+    wedge_quad_order_arbitrary_(q_order)
+{
+  line_quad_order_arbitrary_.SetRange({-1.0, 1.0});
+
+  CreateCellMappings();
+}
```
zachhardy, line 35: Are we initializing all quadratures for all element types here? What is the rationale for this instead of initializing what is needed based on element type?

framework/math/SpatialDiscretization/FiniteElement/Lagrange/LagrangeBase.cc
```diff
@@ -0,0 +1,149 @@
...
+
+void LagrangeBase::CreateCellMappings()
+{
+  typedef cell_mapping::LagrangeSlabMapping SlabSlab;
```
zachhardy: Is `SlabSlab` a typo?

framework/physics/FieldFunction/ff_gridbased_00_constrdestr.cc
```diff
@@ -134,12 +137,25 @@ chi_math::SDMPtr FieldFunctionGridBased::MakeSpatialDiscretization(
...
     chi_math::CoordinateSystemType::CARTESIAN;
+  std::string cs = "cartesian";
+  if (user_params.Has("coordinate_system"))
+  {
+    cs = params.GetParamValue<std::string>("coordinate_system");
+
+    using namespace chi_math;
+    if (cs == "cartesian") cs_type = CoordinateSystemType::CARTESIAN;
+    if (cs == "rz") cs_type = CoordinateSystemType::CYLINDRICAL;
+    if (cs == "1d_spherical") cs_type = CoordinateSystemType::SPHERICAL;
```
zachhardy: It feels icky to me to have coordinate system and dimensionality inherently tied together in the input language for cylindrical and spherical, but not cartesian. It would be could to see `cartesian`, `cylindrical`, and  `spherical` instead with subsequent checks on the mesh dimension for validity.

framework/physics/FieldOperations/PartitionerPredicate.h
```diff
@@ -0,0 +1,31 @@
...
+
+namespace chi
+{
+class GraphPartitioner;
+}
+
+namespace chi_physics::field_operations
+{
+
+class PartitionerPredicate : public FieldOperation
```
zachhardy, line 14: I won't be using this, but I am unsure what "PartitionerPredicate" means. Am I reading the source code correctly when I interpret this as plopping cell processor IDs into a specific component of a specific field function? Is the idea that you could create a mesh and a FF in Lua and then apply this?

zachhardy (approved):
A few minor comments:

1) While most can recognize what `CG` and `DG` are in the test folder structure, we don't really use `CG` or `DG` verbiage in the code. I would encourage using `CFEM` and `DFEM` to be more in line with the naming within the code.

2) A little bit out of scope, but I incidentally got confused by the `CellVolumeIntegralPostProcessor`. I thought it was a post-processor that integrates a single cell. I would recommend renaming this to either `VolumeIntegralPostProcessor` or `IntegralPostProcessor`.

framework/math/SpatialDiscretization/CellMappings/Lagrange/LagrangeHexMapping.h
Naktakala: The override states that this child class's method overrides the virtual method from the parent

framework/math/SpatialDiscretization/FiniteElement/Lagrange/LagrangeBase.cc
Naktakala: Basically, the quadratures can be recycled given the same element type and are fairly simple to create, however, both their storage and initialization can become expensive in numbers. For example, a quadrilateral quadrature has a Legendre quadrature on the inside. Initializing 50,000 legendre quadratures would become noticeable if there are 50,000 quads there. Since they can be recycled and easy to store, we just go ahead and store them.

zachhardy: I wasn't proposing initializing on a per-cell mapping basis, but rather, just not initializing hex quadrature in a slab problem (as an example). I suppose for mixed element type meshes, this can become hairy with checks quite quickly, making initializing everything the cleaner alternative.

framework/math/SpatialDiscretization/FiniteElement/Lagrange/LagrangeBase.cc
Naktakala: Indeed.

framework/physics/FieldFunction/ff_gridbased_00_constrdestr.cc
Naktakala: IDK. I've had the same icky feeling when seeing cartesian (X, XY, XYZ), cylindrical (R, RZ, RZ-Varphi), spherical (R, R-Varphi, R-Varphi-Theta) in other codes. There is always confusion on the dimensionality when one does that. There are few examples of codes that do the 3D cylindrical and 2D+3D spherical variant.  In fact, Kernels, and stuff become substantially more complicated with the cylindrical 3D variants. I particularly put this in here to concisely remind the user that this is the intended dimension. This is what moose has:
```
MultiMooseEnum coord_types("XYZ RZ RSPHERICAL", "XYZ");
```

I did identify now that we can by accident try to make an rz field with 1D cells. Created issue #396 for this.

framework/physics/FieldOperations/PartitionerPredicate.h
Naktakala: This might help a little.
```
params.SetGeneralDescription("Field operation that will write to the "
                               "field the result of a partitioning operation");
```
It has two use cases thus far. Diagnostic when partitioning in serial goes wrong, and if you want to build volumetric sub-sets of cells.

framework/math/SpatialDiscretization/CellMappings/LagrangeBaseMapping.h
Naktakala: I think I am going to stick with World and Natural here to match the textbook I based it off. There are also tons of google-able stuff popping up when I search for something like `quad natural coordinates` as opposed to `quad reference coordinates`. Deal.ii uses RealToUnitCell but I don't really like the "real" part.

I did remove the Lagrange part here to reduce confusion however when we have other methods doing something similar we will have segregate it again.

framework/math/SpatialDiscretization/CellMappings/LagrangeBaseMapping.cc
Naktakala: Done

framework/math/SpatialDiscretization/CellMappings/LagrangeBaseMapping.cc
Naktakala: Done

framework/math/SpatialDiscretization/CellMappings/LagrangeBaseMapping.h
zachhardy: World and Natural are fine as long as it is documented. I would recommend `NaturalXYZ` in the class name as well. I don't know what other methods of mapping coordinates you plan to put in (or if it will be in this class or not), but we'll cross that bridge when we get to it.

framework/physics/FieldFunction/ff_gridbased_00_constrdestr.cc
zachhardy: Maybe you misunderstood my comment. These inputs specify the coordinate system type. Cartesian, cylindrical, and spherical geometries can be 1d, 2d, or 3d. 

I am *not* recommending to use `x`, `xy`, and `xyz`-type options, but rather, dimension independent `cartesian`, `cylindrical`, and `spherical` options. For whatever the user is trying to do, if a particular coordinate system-dimensionality pair is not implemented, the code should puke. 

Why not name the options after the options in the `CoordinateSystemType` enum class? If we had full 3D in all coordinate systems, then if a user specifies `1d_cylindrical`, `rz`, or `3d_cylindrical`, it would result in `cs_type = CoordinateSystemType::CYLINDRICAL`, which makes specifying the coordinate system in a dimensionality-dependent way unnecessary and makes the input language less clear (especially when considering 1d, 2d, and 3d Cartesian is just `cartesian`).

PR #398, opened 2023-10-27
Changes 2023 10 week4

Naktakala:
## Highlights
- Added some lua functions so that users can use timing blocks from input language
- Reworked element volume and face-area computations to address issue #397 

## Notes
The cell mapping stuff might look intimidating because of all the files changed but the changes were fairly simple. None of the API's changed. Also, the computation of element-volumes and areas are now done via an Initialize method which does not follow RAII, however, I had a hard time figuring out a solution and this is the only one I could come up with.

### DIFF
--- a/framework/math/SpatialDiscretization/CellMappings/PieceWiseLinear/pwl_slab_00_constrdestr.cc
+++ b/framework/math/SpatialDiscretization/CellMappings/PieceWiseLinear/pwl_slab_00_constrdestr.cc
@@ -1,6 +1,11 @@
 #include "PieceWiseLinearSlabMapping.h"
+
 #include "mesh/MeshContinuum/chi_meshcontinuum.h"
+
+#include "math/SpatialDiscretization/SpatialDiscretization.h"
 #include "math/SpatialDiscretization/CellMappings/PieceWiseLinearBaseMapping.h"
+
+#include "chi_log.h"
 
 namespace chi_math::cell_mapping
 {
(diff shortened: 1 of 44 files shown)


framework/math/SpatialDiscretization/CellMappings/PieceWiseLinear/pwl_slab_00_constrdestr.cc
```diff
@@ -1,30 +1,69 @@
...
 
-  chi_mesh::Vector3 v01 = v1 - v0_;
+  chi_mesh::Vector3 v01 = v1_ - v0_;
   h_ = v01.Norm();
 
   normals_[0] = slab_cell.faces_[0].normal_;
   normals_[1] = slab_cell.faces_[1].normal_;
 }
 
+void PieceWiseLinearSlabMapping::ComputeCellVolumeAndAreas(
```
ragusa, line 37: what is piecewiselinearslab? is it just a linearslab?

ragusa (commented):
I specifcially looked in the commit Corrected incorrect volume computations of FV mappings to see what was wrong. it's a long commit. What was the nature of the bug exactly for FV-rz?

Naktakala:
> I specifcially looked in the commit Corrected incorrect volume computations of FV mappings to see what was wrong. it's a long commit. What was the nature of the bug exactly for FV-rz?

The volume would just always be the cartesian volume.

framework/math/SpatialDiscretization/CellMappings/PieceWiseLinear/pwl_slab_00_constrdestr.cc
Naktakala: Yes. And now that we have lagrange SDM we can probably remove it. Wanna volunteer?

PR #399, opened 2023-11-09
Add missing headers

loganharbour:
These are required for modern clang builds.


loganharbour:
Looks like @Naktakala made these changes a few weeks ago :)

